PR #3, opened 2017-12-28
Converge stylesheets

plicjo:


app/assets/stylesheets/rad_common/radbear_mailer.css
```diff
@@ -32,10 +32,6 @@
     margin-right: 40px;
 }
 
-.text-right {
```
plicjo: This is available in Bootstrap.

PR #14, opened 2018-03-22
Update to the latest version of autocomplete

plicjo:
Updating to the most current version of the rails-autocomplete javascript.

https://github.com/bigtunacan/rails-jquery-autocomplete/blob/master/lib/assets/javascripts/autocomplete-rails-uncompressed.js

### DIFF
--- a/app/assets/javascripts/rad_common/autocomplete-rails.js
+++ b/app/assets/javascripts/rad_common/autocomplete-rails.js
@@ -17,14 +17,17 @@
 (function(jQuery)
 {
   var self = null;
-  jQuery.fn.railsAutocomplete = function() {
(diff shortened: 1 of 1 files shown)


plicjo:
@murjax Do you think this PR will affect the global search results in Groundswell or Smart Pharmacy?

Didn't you override some of this library's functionality in those two apps?

plicjo:
Important: I've got it working fine on a branch in Groundswell.

murjax:
@plicjo Yes, our current autocomplete-rails.js has custom code that does two things:

1. Show detailed results for task searches in GS
2. Trigger a change event when user selects prescriber location on order form in SP. 

We should try to extract these features. I remember having problems intercepting the process outside of autocomplete, but it's worth another shot.

PR #16, opened 2018-04-06
Show Data Columns Field Translation

murjax:
Check locale translations for field name. If one exists, show that. Otherwise show the field name as is.


garyfoster:
Nice!! I think since this could affect a lot of stuff, we should add one more test assertion. Assert that a non translated value still works with the title case.

PR #18, opened 2018-04-11
Allow multiple security role selection (prototype)

murjax:
This displays the collection of security roles as checkboxes. The form is loaded with the user's security roles checked. When a role is selected and posted, the controller will parse the role ids and set the user's security groups accordingly.

### DIFF
--- a/app/helpers/rad_common/security_role_helper.rb
+++ b/app/helpers/rad_common/security_role_helper.rb
@@ -31,5 +31,9 @@ module RadCommon
         label
       end
     end
+
+    def collection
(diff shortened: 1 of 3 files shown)


app/helpers/rad_common/security_role_helper.rb
```diff
@@ -31,5 +31,9 @@ def normalize_name(label)
         label
       end
     end
+
+    def collection
```
garyfoster: we should probably name this something since a  helper is in the global scope, `collection` is very general

PR #21, opened 2018-04-12
User Security Role Tests

murjax:
- Add tests to confirm user's `updated_at` is touched when security roles are added or removed.
- Changed the `updated_at` touch functionality to use `before_create` and `after_destroy` callbacks.

### DIFF
--- a/app/models/concerns/radbear_user.rb
+++ b/app/models/concerns/radbear_user.rb
@@ -50,7 +50,7 @@ module RadbearUser
     end
 
     if response.body && response.body.count != 0
-      response.body || []
+      response.body
     else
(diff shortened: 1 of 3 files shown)


app/models/concerns/radbear_user.rb
```diff
@@ -50,7 +50,7 @@ def firebase_device_tokens(app)
     end
 
     if response.body && response.body.count != 0
-      response.body || []
+      response.body
```
murjax: Found this while messing with the callbacks. That statement can never return an empty array on that line because you've already confirmed the presence of `response.body`

PR #54, opened 2019-02-04
Model link fixes

elbzero:


spec/dummy/app/models/status.rb
```diff
@@ -0,0 +1,9 @@
+class Status < ApplicationRecord
+  include Authority::Abilities
+  audited
+
+  def to_s
```
garyfoster: use alias_attribute

spec/dummy/db/migrate/20190204202421_create_statuses.rb
```diff
@@ -0,0 +1,9 @@
+class CreateStatuses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :statuses do |t|
+      t.string :name
```
garyfoster: make not nullable and unique just so it's more realistic

PR #57, opened 2019-03-21
Simple autocomplete

elbzero:


app/assets/javascripts/rad_common/autocomplete-rails.js
```diff
@@ -45,17 +45,18 @@
   {
       jQuery(".ui-autocomplete-input").each( function()
                                              {
-                                              display_autocomplete_error(this);
+                                              display_autocomplete_error(this, "id-element");
+                                              display_autocomplete_error(this, "association-element");
```
elbzero: Display errors for either the associated element or the id element depending on where there error was added in the model

app/views/layouts/_simple_autocomplete.html.haml
```diff
@@ -0,0 +1,7 @@
+= f.input "#{association.to_s}_id", as: :hidden
```
elbzero: adding the id field here so we don't have to do it manually

app/views/layouts/_simple_autocomplete.html.haml
```diff
@@ -0,0 +1,7 @@
...
+= f.input association, as: :hidden
```
elbzero: adding the associated field here so that if we get any errors on the association we can pull them up to the autocomplete field

PR #59, opened 2019-03-23
testing notify

calebtocco:


app/controllers/notification_settings_controller.rb
```diff
@@ -9,7 +9,7 @@ def index
     @notification_settings = []
 
     notifications.each do |notification|
-      @notification_settings.push NotificationSetting.find_or_initialize_by(notification_type: notification)
+      @notification_settings.push NotificationSetting.find_or_initialize_by(notification_type: notification, user_id: current_user.id)
```
garyfoster: this fix is good for sure

app/views/notification_settings/index.html.haml
```diff
@@ -16,7 +16,7 @@
               = simple_form_for(notification_setting) do |f|
                 = f.error_notification
                 .form-inputs
-                  = f.input :enabled
+                  = f.input :enabled, input_html: { id: "notification_setting_enabled_#{notification_setting.notification.description.parameterize.underscore}" }
```
garyfoster: curious about what is going on here, do you think it's still needed?

app/controllers/notification_settings_controller.rb
garyfoster: would be great to add a test for this in rad_common in notification_settings_controller_spec

garyfoster: actually, since we are setting that user in the create, is that needed? we could put the user attribute in the form as hidden if that helps, then we wouldn't have to set it to current_user in the create

app/views/notification_settings/index.html.haml
```diff
@@ -13,7 +13,7 @@
           %td= notification_setting.notification.description
           %td
             -if can_update && current_user.can_update?(notification_setting)
-              = simple_form_for(notification_setting) do |f|
+              = simple_form_for notification_setting, url: '/notification_settings', method: :post do |f|
```
garyfoster: a cleaner way is to use the helper like `notifications_path` fyi - i'll fix

PR #70, opened 2019-05-29
Bootstrap4

nativestranger:


Gemfile.lock
```diff
@@ -14,6 +14,7 @@ PATH
       audited (~> 4.8)
       authority (= 3.3.0)
       aws-sdk-s3
+      bootstrap (~> 4.3.1)
```
garyfoster: nice, this looks like a much better and more well maintained gem

PR #72, opened 2019-05-31
Bootstrap4

nativestranger:


Gemfile
```diff
@@ -16,7 +16,9 @@ gemspec
 gem 'schema_validations',  git: 'https://github.com/SchemaPlus/schema_validations.git'
 
 group :test do
-  gem 'capybara-webkit'
+  gem 'capybara'
+  gem 'capybara-selenium'
```
garyfoster: What's driving these changes?

app/assets/stylesheets/rad_common/rad_common.css
```diff
@@ -259,3 +268,92 @@ td.fa.security-group-td { border-style: none; }
     margin-bottom: 0;
   }
 }
+
```
garyfoster: Enhance the comments on this section so we have more insights when looking back.

app/helpers/rad_common/bootstrap_flash_helper.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module RadCommon
+  module BootstrapFlashHelper
```
garyfoster: What's the story with this helper?

spec/dummy/app/assets/stylesheets/bootstrap_and_overrides.scss
```diff
@@ -0,0 +1,7 @@
+html {
```
garyfoster: What are the items in this file used for?

spec/rails_helper.rb
```diff
@@ -86,4 +86,23 @@
   end
 end
 
-Capybara.javascript_driver = :webkit
+Capybara.register_driver :headless_chrome do |app|
```
garyfoster: @elbzero is working on some overall improvements to the test suite so we need to coordinate with him. If this stuff isn't needed for the bootstrap change, let's separate this out so it can be worked on independently.

garyfoster:
This is looking really good. See other comments inline as well as these here:

1. The sign in / sign out is on the left on the top nav instead of the right - I don't hate it, was that by design?
2. The checkbox looks funky on the sign in form, let me know if you don't see it.
3. The global search form field with the prefix and suffix add-ons look bad. The checkbox looks funky there too.

app/views/layouts/_navigation_admin.html.haml
```diff
@@ -2,13 +2,13 @@
...
-  %li.divider
-  %li= link_to 'Users', users_path
-  %li= link_to 'Security Roles', security_roles_path
-  %li= link_to 'Notifications', '/rad_common/notification_types'
-  %li= link_to 'Company Info', edit_company_path(company)
-  %li= link_to 'System Usage', '/rad_common/system_usages'
-  %li= link_to 'System Message', RadCommon::Engine.routes.url_helpers.new_system_message_path
-  %li= link_to 'Audit Search', audit_search_users_path
+  %li.dropdown-divider
+  %li= link_to 'Users', users_path, class: 'dropdown-item'
```
garyfoster: Why do we have to add this class dropdown-item?

Gemfile
nativestranger: capybara-webkit is no longer supported and can't be installed on newer version of macOS (like the one I'm running). I was unable to work on the gem without making these changes. This now uses the same gems for browser tests that tracker does

garyfoster: Ok, I wasn't aware of that. It's working for the rest of us on our dev machines, but we'll start the process of switching over to the new gems. I do remember it being a pain to install the native libraries. I might pull this out separately ahead of time depending on timing.

garyfoster: @elbzero see above

app/views/layouts/_navigation_admin.html.haml
nativestranger: https://getbootstrap.com/docs/4.0/migration/#dropdowns

"Dropdown items now require .dropdown-item" for styling

app/assets/stylesheets/rad_common/rad_common.css
```diff
@@ -259,3 +267,62 @@ td.fa.security-group-td { border-style: none; }
...
+    white-space: nowrap;
+  }
+  .dl-horizontal dd {
+    margin-left: 180px;
+    margin-bottom: 0;
+  }
+}
+
+.dl-horizontal dd:before,
+.dl-horizontal dd:after, {
```
garyfoster: My IDE doesn't like this comma, is it valid?

app/helpers/rad_common/bootstrap_flash_helper.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
```
garyfoster: What does this comment mean?

app/helpers/rad_common/bootstrap_flash_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module RadCommon
+  module BootstrapFlashHelper
+    ALERT_TYPES = %i[success info warning danger].freeze unless const_defined?(:ALERT_TYPES)
+
+    # rubocop:disable all
```
garyfoster: Why are we disabling rubocop here?

spec/dummy/app/assets/stylesheets/bootstrap_and_overrides.scss
garyfoster: ?

spec/dummy/config/initializers/rad_common.rb
```diff
@@ -49,3 +49,5 @@
       query_where: 'divisions.name ilike :search',
       super_search_exclude: true }
   ]
+
+ActionController::Base.send :helper, RadCommon::BootstrapFlashHelper
```
garyfoster: This looks weird, was there a problem with this helper being accessible?

garyfoster:
We're down to just very minor things here, looking great.

garyfoster:
Is the sign in / sign out nav supposed to be moved from the right to the left? By design or bug?

garyfoster:
The checkbox looks funky on the sign in form on the demo app, looks fixed on the CSA project

spec/dummy/app/assets/stylesheets/bootstrap_and_overrides.scss
nativestranger: This is just for importing bootstrap and setting default overriding values

app/helpers/rad_common/bootstrap_flash_helper.rb
nativestranger: Ruby 3 was going to make string literals frozen by default (until Matz decided not to as of 6 months ago, apparently). This 'magic comment' let's us opt into the default frozen string literal functionality on a per-file basis

app/helpers/rad_common/bootstrap_flash_helper.rb
nativestranger: Just because this code was pulled from a battle-tested gem, and rubocop was complaining about a bunch of stuff that I figured we wouldn't care to fix right now

spec/dummy/config/initializers/rad_common.rb
nativestranger: Yea. The helper isn't included by default. We could have the module extend itself and then call the helper function on the module itself in the layout. But that also is not commonly seen in a rails developer's perspective.

nativestranger:
> The checkbox looks funky on the sign in form on the demo app, looks fixed on the CSA project

fixed

nativestranger:
> Is the sign in / sign out nav supposed to be moved from the right to the left? By design or bug?

That was by accident. Think it's worth fixing?

app/helpers/rad_common/bootstrap_flash_helper.rb
garyfoster: ok

spec/dummy/app/assets/stylesheets/bootstrap_and_overrides.scss
```diff
@@ -0,0 +1,7 @@
...
+  font-size: 14px;
+}
+
+// Custom bootstrap variables must be set or imported *before* bootstrap.
+$secondary: #f5f5f5;
```
garyfoster: Why override the secondary button color, if we use btn-light it looks like the same as this overridden one.

app/views/layouts/_navigation_right.html.haml
```diff
@@ -1,16 +1,15 @@
 %ul.nav.navbar-nav.navbar-right
```
garyfoster: This nav should be on the right.

spec/dummy/config/initializers/rad_common.rb
garyfoster: I removed this and it seems to work, please confirm.

spec/dummy/app/assets/stylesheets/bootstrap_and_overrides.scss
nativestranger: It does look very similar. I decided to not override for now

app/assets/stylesheets/rad_common/rad_common.css
```diff
@@ -259,3 +267,62 @@ td.fa.security-group-td { border-style: none; }
...
+}
+
+/* preserved from bootstrap 3 panel-default styles */
+/* TODO: convert views to standard card usage and remove the below */
+.card-default {
+  border-color: #ddd;
+}
+.card-default > .card-header {
+  color: #333;
+  background-color: #fff;
```
elbzero: Just curious, are we changing the color here? Just saw this as I was implementing this for better_way_ars. Our old one was more of a grey. No problem either way, just making sure this is intentional

PR #77, opened 2019-06-18
Active storage migration rake

calebtocco:


app/services/migrate_paperclip_data.rb
```diff
@@ -6,17 +6,17 @@ class MigratePaperclipData
   attr_accessor :new_attachment_name
   attr_accessor :model_class
 
-  def self.perform(model_class, attachments_to_process)
+  def self.perform(model_class, attachment_names)
```
calebtocco: switched these so they derive the new attachment names, should make passing args to the rake task more trivial

app/services/migrate_paperclip_data.rb
```diff
@@ -38,7 +38,14 @@ def perform_migration
 
     ActiveRecord::Base.transaction do
       model_class.where("#{attachment_file_name} is not null").each do |record|
+        if record.send(new_attachment_name).attached?
```
garyfoster: In what scenario could this ever occur?

PR #78, opened 2019-06-20
multiple models and attachments

calebtocco:


app/services/migrate_paperclip_data.rb
```diff
@@ -18,11 +17,12 @@ def self.perform(model_class, attachment_names)
       migrator.attachment_file_size = "#{attachment_name}_file_size"
       migrator.new_attachment_name = "#{attachment_name}_new"
 
+      migrator.prepare_sql_statements unless sql_prepared
```
calebtocco: it was crashing when attempting to prepare the sql the second time through

PR #121, opened 2019-08-20
attachment layout

calebtocco:


app/assets/stylesheets/rad_common/rad_common.scss
```diff
@@ -185,6 +185,12 @@ td.fa.security-group-td { border-style: none; }
   text-align: center;
 }
 
+.img-thumbnail {
```
garyfoster: is there a default bootstrap class/component we can use instead of custom CSS>

PR #124, opened 2019-08-29
twilio system messages

calebtocco:


app/jobs/system_sms_job.rb
```diff
@@ -0,0 +1,23 @@
...
+          RadicalTwilio.send_sms(to: phone_number, message: message, from: RadicalTwilio.twilio_phone_number)
+        end
+      rescue Twilio::REST::RequestError
+        errors << user.id
+      end
+    end
+
+    return if errors.empty?
+
+    body = "These users did not receive a system SMS message: #{errors.join(', ')}"
```
calebtocco: just shows user ids, could use improvement

app/validators/twilio_phone_validator.rb
```diff
@@ -0,0 +1,43 @@
+class TwilioPhoneValidator < ActiveModel::Validator
```
calebtocco: extracted from SP

app/services/radical_twilio.rb
```diff
@@ -0,0 +1,17 @@
+class RadicalTwilio
```
calebtocco: extracted from SP

app/models/concerns/rad_company.rb
```diff
@@ -72,6 +75,34 @@ def usage_stats
     [usage_headers, usage_items, usage_data]
   end
 
+  def next_phone_number
```
calebtocco: extracted from SP

app/controllers/system_messages_controller.rb
```diff
@@ -15,20 +17,28 @@ def create
...
     if @system_message.save
-      @system_message.send!
+      @system_message.send!(current_user)
       redirect_to "/rad_common/system_messages/#{@system_message.id}", notice: 'The message was successfully sent.'
     else
       render :new
     end
   end
 
+  def twilio
```
garyfoster: Is this needed? It may just be used for robo calls?

app/models/system_message.rb
```diff
@@ -29,9 +46,19 @@ def recipients
...
-  def send!
-    recipients.each do |user|
-      RadbearMailer.simple_message(user, "Important Message From #{I18n.t(:app_name)}", message).deliver_later
+  def send!(current_user)
+    if email?
+      recipients.each do |user|
+        RadbearMailer.simple_message(user, "Important Message From #{I18n.t(:app_name)}", message).deliver_later
+      end
+    else
+      SystemSMSJob.perform_later(message, recipients.map(&:id), current_user)
```
garyfoster: If not already, make sure recipients have a phone_number - or mobile_phone, need to see what attribute the other features are using

config/routes.rb
```diff
@@ -8,7 +8,10 @@
     post :global_validity_check, on: :member
   end
 
-  resources :system_messages, only: %i[new create show]
+  resources :system_messages, only: %i[new create show] do
+    post 'twilio', format: 'xml', on: :member
```
garyfoster: I think this is just used for robo calls, please check

app/controllers/system_messages_controller.rb
calebtocco: removed

config/routes.rb
calebtocco: removed

app/controllers/system_messages_controller.rb
```diff
@@ -2,6 +2,8 @@ class SystemMessagesController < ApplicationController
   before_action :authenticate_user!
   before_action :set_system_message, only: %i[show]
 
+  ensure_authorization_performed
```
garyfoster: Was this needed for some reason? If so, we might need it other places.

app/controllers/system_messages_controller.rb
```diff
@@ -15,7 +17,7 @@ def create
     @system_message.user = current_user
 
     if @system_message.save
-      @system_message.send!
+      @system_message.send!(current_user)
```
garyfoster: Since we store the user in the SystemMessage model instance, do we need to pass this?

app/jobs/system_sms_job.rb
```diff
@@ -0,0 +1,33 @@
+class SystemSMSJob < ApplicationJob
+  queue_as :default
+
+  def perform(message, recipients, current_user)
+    errors = []
+    recipients.each do |user_id|
+      user = User.find(user_id)
+      phone_number = if user.mobile_phone.present?
```
garyfoster: What did you find on other usages about phone number? We may want to only use mobile_phone.

app/jobs/system_sms_job.rb
```diff
@@ -0,0 +1,33 @@
...
+        end
+      rescue Twilio::REST::RequestError
+        errors << user.id
+      end
+    end
+
+    return if errors.empty?
+
+    body = "These users did not receive a system SMS message: #{errors.join(', ')}"
+    RadbearMailer.simple_message([current_user.id], 'Error Sending System Message to Some Users', body).deliver_later
```
garyfoster: You can just pass current_user, doesn't need to be an array, or an ID (i think)

app/models/concerns/rad_company.rb
garyfoster: yep - FYI, in case you were wondering, it cycles from available phone numbers to avoid a limit on sending too many text messages from one number

app/models/system_message.rb
garyfoster: did this get done? I don't see it at first glance

app/services/radical_twilio.rb
```diff
@@ -0,0 +1,17 @@
...
+  def self.send_robocall(from:, to:, url:)
+    client.calls.create(from: from, to: to, url: URI::encode(url))
+  end
+
+  def self.client
+    Twilio::REST::Client.new
+  end
+
+  def self.twilio_phone_number
+    Company.main.next_phone_number
```
garyfoster: I'm thinking that next_phone_number logic/method should be moved out of the company model into this class where it seems to belong more

spec/dummy/db/migrate/20190829220515_add_message_type_to_system_messages.rb
```diff
@@ -0,0 +1,7 @@
+class AddMessageTypeToSystemMessages < ActiveRecord::Migration[5.2]
+  def change
+    add_column :system_messages, :message_type, :integer, null: false
```
garyfoster: existing system messages in production will cause this to crash since they have nils

garyfoster: nevermind, this is RC

spec/dummy/db/migrate/20190829220515_add_message_type_to_system_messages.rb
```diff
@@ -0,0 +1,7 @@
+class AddMessageTypeToSystemMessages < ActiveRecord::Migration[5.2]
```
garyfoster: This migration should be copied in the install_generator

app/models/system_message.rb
calebtocco: yep, performed the check in the job

app/controllers/system_messages_controller.rb
calebtocco: removed

app/models/system_message.rb
garyfoster: I mean, we should scope the recipients query to only include users that have the appropriate phone number present

app/controllers/system_messages_controller.rb
calebtocco: nope removed

app/jobs/system_sms_job.rb
calebtocco: i misunderstood, mobile_phone should be all thats needed

app/services/radical_twilio.rb
calebtocco: that method updates the specific instance (https://github.com/radicalbear/rad_common/pull/124/files#diff-9880236d7a1c3cf97ee0e0a023a4713bR92), although i think its safe to say Company.main will always be used

app/services/radical_twilio.rb
```diff
@@ -0,0 +1,40 @@
...
+  def self.send_robocall(from:, to:, url:)
+    client.calls.create(from: from, to: to, url: URI::encode(url))
+  end
+
+  def self.client
+    Twilio::REST::Client.new
+  end
+
+  def self.twilio_enabled?
+    defined?(Twilio).present? && Company.has_attribute?(:twilio_phone_numbers) && client.present?
```
garyfoster: I was thinking we could just check if the env vars exist, or was this better for some reason?

app/models/concerns/rad_company.rb
```diff
@@ -8,12 +8,15 @@ module RadCompany
...
-    schema_validations except: :valid_user_domains
+    schema_validations except: %i[twilio_phone_numbers valid_user_domains]
+
+    before_validation :sanitize_twilio_numbers
 
     validates_with EmailAddressValidator, fields: %i[email]
     validates_with PhoneNumberValidator
     validate :validate_only_one, on: :create
     validate :validate_domains
+    validates :twilio_phone_numbers, presence: true, if: -> { RadicalTwilio.twilio_enabled? && Company.has_attribute?(:twilio_phone_numbers) }
```
garyfoster: In hindsight, I think we should move twilio_phone_numbers from the database and put into an env var, and if only a single phone number is needed, the format would be the same, there would just be one

app/models/concerns/rad_company.rb
```diff
@@ -72,6 +79,12 @@ def usage_stats
     [usage_headers, usage_items, usage_data]
   end
 
+  def sanitize_twilio_numbers
+    return unless RadicalTwilio.twilio_enabled? && Company.has_attribute?(:twilio_phone_numbers)
```
garyfoster: all logic involving twilio_phone_numbers would be simplified, any twilio app would have the required env var

PR #127, opened 2019-09-05
Background devise mailers to active job

garyfoster:


garyfoster:
This won't work, have to override it in the user.rb model

PR #145, opened 2019-10-01
attachment deletion and upgrades

calebtocco:
https://swell.radicalbear.com/tasks/30620

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -1,0 +1,14 @@
+class AttachmentsController < ApplicationController
+  before_action :authenticate_user!
+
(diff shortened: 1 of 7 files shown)


app/controllers/attachments_controller.rb
```diff
@@ -0,0 +1,14 @@
+class AttachmentsController < ApplicationController
+  before_action :authenticate_user!
+
+  authority_actions destroy: 'update'
+
+  def destroy
+    attachment = ActiveStorage::Attachment.find(params[:id])
+    record = attachment.record
+    authorize_action_for record
```
garyfoster: slick, I like it

PR #146, opened 2019-10-03
permanent_attachment_urls

calebtocco:
https://swell.radicalbear.com/tasks/30630


app/models/concerns/attachable.rb
```diff
@@ -0,0 +1,9 @@
+module Attachable
+  include Hashable
+
+  def encoded_url(variant)
```
garyfoster: rename to permanent_attachment_url since that properly identifies the functionality

app/services/attachment_url_generator.rb
```diff
@@ -0,0 +1,9 @@
+class AttachmentUrlGenerator
+  def self.permanent_attachment_variant_url(record, variant)
+    protocol = Rails.env.production? ? 'https' : 'http'
+    host = Rails.env.production? ? ENV.fetch('HOST_NAME') : 'localhost:3000'
```
garyfoster: i've been changing these to require HOST_NAME locally

PR #155, opened 2019-11-05
Add filtering patterns

elbzero:


app/helpers/rad_common/application_helper.rb
```diff
@@ -80,12 +80,13 @@ def enum_to_translated_option(klass, enum, enum_value, default = enum_value.to_s
       I18n.t(key, default: default)
     end
 
-    def options_for_enum(klass, enum)
+    def options_for_enum(klass, enum, db_values=false)
```
garyfoster: I feel like we should add a different method for db values rather than using the same method with a switch but I don't fully know what it's used for yet

garyfoster:
Before I go through this in more detail, are there any methods in the various classes that should be made private, not required in the public API?

elbzero:
> Before I go through this in more detail, are there any methods in the various classes that should be made private, not required in the public API?

Updated what I could see for these, might do another pass over later, but should be good for now

app/helpers/rad_common/application_helper.rb
elbzero: updated per our Slack conversation

app/services/rad_common/date_filter.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def filter_view
+      'date'
+    end
+
+    def searchable_name
+      [start_input, end_input]
+    end
+
+    def start_at_value(params)
```
garyfoster: Are start_at_value and end_at_value needing to be public? I can't easily tell. And apply_filter?

app/services/rad_common/filter_defaulting.rb
```diff
@@ -0,0 +1,45 @@
+module RadCommon
+  module FilterDefaulting
```
garyfoster: for non-class modules, my preference is to not create a separate file simply to break it up

app/services/rad_common/filtering.rb
```diff
@@ -0,0 +1,81 @@
+module RadCommon
```
garyfoster: another non-class module, same issue

app/services/rad_common/like_filter.rb
```diff
@@ -0,0 +1,31 @@
+module RadCommon
+  class LikeFilter
```
garyfoster: classes like this make sense to me to be separate

app/services/rad_common/like_filter.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def filter_view
+      'like'
+    end
+
+    def searchable_name
+      like_input
+    end
+
+    def like_value(params)
```
garyfoster: can't tell if this should be public, maybe take another pass - being able to see private/public is important when trying to understand the code when debugging and adding features

app/services/rad_common/sorting.rb
```diff
@@ -0,0 +1,48 @@
+module RadCommon
+  module Sorting
```
garyfoster: another module, my pref is to not split these out

app/views/layouts/_filter_select.html.haml
```diff
@@ -0,0 +1,7 @@
+= f.input filter.searchable_name, collection: filter.input_options, label: filter.input_label,
+                                  selected: search.selected_value(filter.searchable_name), include_blank: filter.blank_value_label,
+                                  as: filter.input_type, group_method: :last, required: false, label_method: filter.label_method,
+                                  input_html: { class: 'selectpicker',
+                                                multiple: filter.multiple,
+                                                'data-live-search' => true,
```
garyfoster: can we correct the hash syntax? this should work?: `'data-live-search': true`

garyfoster: there are more instances of the hash syntax, pls do a search

app/services/rad_common/date_filter.rb
elbzero: Fixing, start_at_value and end_at_value are private but apply_filter needs to be public

app/services/rad_common/like_filter.rb
elbzero: agreed, like_value does need to be private

app/views/layouts/_filter_select.html.haml
elbzero: fixed

app/services/rad_common/filter_defaulting.rb
elbzero: Yeah I was kind of back and forth on this pattern too. My main issue was that I was stuffing too much logic into a single class and making it concerned with multiple things which causes too much complexity. I'm going to try to quick refactor and break these out into classes instead. I am putting in a separate branch so that if causes too much trouble we can abort easily

app/views/layouts/_sort_search_headers.html.haml
```diff
@@ -0,0 +1,16 @@
+- unless defined?(sort_columns)
```
garyfoster: thought we were going to name this _sort_headers and then name the old ones _legacy_sort_headers?

elbzero: Just did this now. I updated in the hmpbrands branch as well. Though we will need to update on each project as we go

PR #169, opened 2019-11-22
Refactor app info for list of models and tables

garyfoster:
Use a more robust and unified method to return application information such as the list of tables and models.


app/services/global_validity.rb
```diff
@@ -69,25 +69,23 @@ def needs_to_run?
     def models_to_check
       return [@override_model.constantize] if @override_model.present?
 
-      Rails.application.eager_load!
```
garyfoster: removed this eager loading stuff which seems sketchy

PR #174, opened 2019-11-30
skippable index schema validations

calebtocco:
https://swell.radicalbear.com/tasks/31248

### DIFF
--- a/spec/dummy/app/models/division.rb
+++ b/spec/dummy/app/models/division.rb
@@ -17,7 +17,7 @@ class Division < ApplicationRecord
   scope :authorized, lambda { |user| where(owner_id: user.id) unless user.admin? }
   scope :sorted, -> { order(:name) }
 
-  validates :name, uniqueness: true, if: -> { status_pending? }
+  validates :name, uniqueness: { message: 'has already been taken for a pending division' }, if: -> { status_pending? }
 
   audited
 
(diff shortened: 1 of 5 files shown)


spec/models/concerns/schema_validations_spec.rb
```diff
@@ -50,13 +50,29 @@
...
+      let(:division) { create :division, division_status: 'status_pending' }
+      let(:division2) { create :division, division_status: 'status_pending' }
+      let(:division3) { create :division, division_status: 'status_active' }
+
+      before do
+        division2.update(name: division.name)
+        division3.update(name: division.name)
+      end
+
+      it 'does not add validation based on schema' do
```
calebtocco: this is a very specific test, which fails if index is not skipped because the where clause in the index would not run otherwise causing the division to be invalid

spec/dummy/app/models/division.rb
```diff
@@ -16,6 +17,8 @@ class Division < ApplicationRecord
   scope :authorized, lambda { |user| where(owner_id: user.id) unless user.admin? }
   scope :sorted, -> { order(:name) }
 
+  validates :name, uniqueness: { message: 'has already been taken for a pending division' }, if: -> { status_pending? }
```
calebtocco: specific message on this for testing

PR #179, opened 2019-12-11
Indicate staging system on email subject

garyfoster:


garyfoster:
So, if I code this differently, I can spec it, if I put the conditional inside the class in the delivering_mail method and mock the env var; but I like it this way better as it doesn't affect production at all, but can't test it. I think writing it this way outweighs the benefits of adding a spec. I found some possible ways to test the initializer that involved some more invasive measures, it just didn't seem worth it.

PR #181, opened 2019-12-13
schema validation loading tests

calebtocco:
https://swell.radicalbear.com/tasks/31299

### DIFF
--- a/lib/core_extensions/active_record/base/schema_validations.rb
+++ b/lib/core_extensions/active_record/base/schema_validations.rb
@@ -1,0 +1,23 @@
+module CoreExtenions
+  module ActiveRecord
+    module Base
+      module SchemaValidations
+        def self.prepended(base)
+          base.singleton_class.send(:prepend, ClassMethods)
+        end
+
+        module ClassMethods
+          def validators
+            new.load_schema_validations unless schema_validations_loaded
+            super
+          end
+
+          def validators_on(*args)
+            new.load_schema_validations unless schema_validations_loaded
+            super
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


spec/dummy/app/models/application_record.rb
```diff
@@ -3,6 +3,6 @@ class ApplicationRecord < ActiveRecord::Base
 
   class_attribute :schema_validations_loaded
 
-  validate :load_schema_validations
+  before_validation :load_schema_validations
```
calebtocco: Adjusting this to match other apps, may change with solution

lib/core_extensions/active_record/base/schema_validations.rb
```diff
@@ -0,0 +1,23 @@
+module CoreExtenions
```
calebtocco: 🐵 🐒

spec/dummy/app/models/application_record.rb
calebtocco: this is still needed in case console is used before a form is visited

spec/dummy/config/initializers/schema_validations.rb
```diff
@@ -0,0 +1,2 @@
+require "#{Gem::Specification.find_by_name('rad_common').gem_dir}/lib/core_extensions/active_record/base/schema_validations"
+ActiveRecord::Base.prepend CoreExtenions::ActiveRecord::Base::SchemaValidations
```
calebtocco: needed in all apps to patch active record

spec/system/schema_validations_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'rails_helper'
```
calebtocco: these will fail if you comment out the initializer

spec/dummy/config/initializers/schema_validations.rb
garyfoster: We now have app/initializers/rad_common.rb for things that need to be in all apps, can you see if this works if the code is moved into there?

calebtocco: moved, seems to work just the same

PR #182, opened 2019-12-16
Twilio 5 update

garyfoster:


app/services/radical_retry.rb
```diff
@@ -3,7 +3,7 @@ class << self
     def perform_request(no_delay: false, &block)
       retries ||= 5
       block.call
-    rescue Net::OpenTimeout, OpenURI::HTTPError, HTTPClient::ConnectTimeoutError, Errno::EPIPE, SocketError, OpenSSL::SSL::SSLError, Errno::ENOENT, Twilio::REST::ServerError => e
+    rescue Net::OpenTimeout, OpenURI::HTTPError, HTTPClient::ConnectTimeoutError, Errno::EPIPE, SocketError, OpenSSL::SSL::SSLError, Errno::ENOENT => e
```
garyfoster: I'm not sure what will happen here with Twilio's new API so removing and we'll wait and see. @calebtocco you'll most likely be the one adjusting this if needed:)

PR #184, opened 2019-12-23
fix resetdb schema validations error

calebtocco:
https://swell.radicalbear.com/tasks/31321

### DIFF
--- a/lib/core_extensions/active_record/base/schema_validations.rb
+++ b/lib/core_extensions/active_record/base/schema_validations.rb
@@ -8,13 +8,21 @@ module CoreExtensions
 
         module ClassMethods
           def validators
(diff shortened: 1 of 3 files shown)


lib/core_extensions/active_record/base/schema_validations.rb
```diff
@@ -8,14 +8,22 @@ def self.prepended(base)
 
         module ClassMethods
           def validators
-            new.load_schema_validations unless schema_validations_loaded
+            new.load_schema_validations if database_exists? && table_exists? && !schema_validations_loaded
```
calebtocco: **table_exists?**  fails unless database exists, so have to check both

PR #195, opened 2020-01-17
refactor initializer config

calebtocco:
https://swell.radicalbear.com/tasks/31490

### DIFF
--- a/lib/rad_common.rb
+++ b/lib/rad_common.rb
@@ -1,6 +1,7 @@
 require 'rad_common/engine'
 
 module RadCommon
+  # Enables/Disables user avatars being uploaded and displayed in the application
   cattr_accessor :use_avatar
 
   def self.setup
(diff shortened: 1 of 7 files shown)


lib/rad_common.rb
```diff
@@ -1,4 +1,11 @@
 require 'rad_common/engine'
 
 module RadCommon
+  # Enables/Disables user avatars being uploaded and displayed in the application
+  cattr_accessor :use_avatar
+  @@use_avatar = false
```
calebtocco: default to false

spec/dummy/config/initializers/rad_common_initializer.rb
```diff
@@ -0,0 +1,3 @@
+RadCommon.setup do |config|
```
garyfoster: this can be moved into the other initializer

PR #198, opened 2020-01-27
WIP More filter pattern updates

elbzero:


spec/system/users_spec.rb
```diff
@@ -92,7 +92,6 @@
         it 'shows users' do
           visit users_path(status: user.user_status_id)
           expect(page).to have_content user.to_s
-          expect(page).to have_content external_user.to_s
```
elbzero: I removed this because previously we were using the same name for each user in the factory but it was switched out to use Faker. The above expectation was a false positive since every user has the same name.

garyfoster: I added it back and made it work. Good catch!

PR #211, opened 2020-02-28
Monkey patch

calebtocco:


garyfoster:
applying this manually in my branch

PR #214, opened 2020-03-05
Notify enhance js

elbzero:


app/assets/javascripts/rad_common/dynamic_updater.js
```diff
@@ -0,0 +1,14 @@
...
+        $.post(url, data)
+            .fail(function(data) {
+                addAlert('danger', data.responseJSON.error)
+            });
+    });
+
+    function addAlert(type, message) {
+        $('.card-body').prepend( $("<div class='alert alert-" + type + "'>" + message + "</div>") )
+    }
+});
```
elbzero: Not sure if you would rather coffee script here but I have kind of been preferring using plain JS instead as coffee script tends to take me more time to write and understand, but I can convert it to coffee if you prefer.

app/assets/javascripts/rad_common/dynamic_updater.js
```diff
@@ -0,0 +1,14 @@
+$(document).ready( function() {
```
elbzero: This is a minimal approach to the pattern we have used in the past to update ajax fields. We can expand on it later to make it work in more situations, but main idea here is not to have to add much extra html and/or controller code for this to work out of the box

app/assets/stylesheets/rad_common/rad_common.scss
```diff
@@ -210,6 +210,12 @@ td.fa.security-group-td { border-style: none; }
   }
 }
 
+.form-inline {
```
garyfoster: Can we limit this to a particular ID in addition to the class? It looks so generic that it could be used elsewhere.

app/controllers/notification_settings_controller.rb
```diff
@@ -19,13 +19,26 @@ def create
...
 
-    result = if notification_setting.save
-               { notice: 'The setting was successfully saved.' }
-             else
-               { error: "The setting could not be saved: #{notification_setting.errors.full_messages.join(', ')}" }
-             end
-
-    redirect_back fallback_location: '/rad_common/notification_settings', flash: result
+    respond_to do |format|
+      format.html do
```
garyfoster: Is the html version still used? It can be removed if not.

app/assets/javascripts/rad_common/dynamic_updater.js
```diff
@@ -0,0 +1,14 @@
...
+    $('form.dynamic-updater input').change(function() {
+        let url = $(this).closest('form').attr('action') + ".json";
+        let data = $(this).closest('form').serializeArray();
+        $.post(url, data)
+            .fail(function(data) {
+                addAlert('danger', data.responseJSON.error)
+            });
+    });
+
+    function addAlert(type, message) {
```
garyfoster: This is kinda funky, the flash messages stay on the page. It's a rare case to get an error, let's just use an alert box and also need to undo the checkbox selection if it fails.

app/assets/stylesheets/rad_common/rad_common.scss
elbzero: agreed, I was going for something more generic at first, but this is becoming more specific so adding a custom class

app/controllers/notification_settings_controller.rb
elbzero: removed the html version

app/assets/javascripts/rad_common/dynamic_updater.js
elbzero: done

PR #217, opened 2020-03-11
use class names for autocomplete instead of ids to prevent js warnings

elbzero:
[Task](https://swell.radicalbear.com/tasks/31897)


garyfoster:
moved here: https://github.com/radicalbear/rad_common/pull/225

PR #227, opened 2020-05-12
start on documenting

elbzero:
[Task](https://swell.radicalbear.com/tasks/31841)


app/controllers/users_controller.rb
```diff
@@ -4,34 +4,36 @@ class UsersController < ApplicationController
...
+    filters = [{ input_label: 'Status',
+                 column: :user_status_id,
+                 options: UserStatus.not_pending.by_id,
+                 default_value: UserStatus.default_active_status.id }]
 
-    @users = policy_scope(User).recent_first
-    @users = @users.where(user_status: @status) if @status
-    @users = @users.where(external: params[:external]) if params[:external].present?
+    if RadCommon.external_users
+      filters.push({ input_label: 'Type', column: :external, scope_values: %i[internal external] })
```
garyfoster, line 17: @elbzero - why is `column` required for this type of search filter?

garyfoster:
@elbzero as part of this I converted the search on the users index page, can you check my work?

app/controllers/users_controller.rb
garyfoster: @elbzero ^

garyfoster:
> @elbzero as part of this I converted the search on the users index page, can you check my work?

@elbzero ^

app/controllers/users_controller.rb
elbzero: @garyfoster I believe its because we had a situation where we wanted to mix scope values with regular values for example: Users Unassigned, John, Joe, so we need to know what column to filter when its not a 'scope value' we can relax that constraint if needed.

elbzero:
> > @elbzero as part of this I converted the search on the users index page, can you check my work?
> 
> @elbzero ^

@garyfoster Reviewed, looks good!

elbzero:
> this file already shows ignored for me, is this needed?

it came up in my git in rubymine so not sure why it doesn't show on yours?

PR #230, opened 2020-05-18
create tests for email and password changed notifications.

Jgerena11:
https://swell.radicalbear.com/tasks/32430


garyfoster:
good @Jgerena11!

PR #243, opened 2020-06-15
allow html 5 date pickers with data-html-5 = true

elbzero:


app/assets/javascripts/bootstrap_datetimepicker/dates.coffee
```diff
@@ -96,7 +96,7 @@ class DateTimePicker
     format = 'YYYY-MM-DD'
     view = 'M/D/YYYY'
     # Update Values
-    sel = $("input.date")
+    sel = $("input.date:not([data-html-5='true'])")
```
elbzero, line 99: This is opt in so making sure not to enable the js or css when html 5 is enabled

PR #254, opened 2020-08-09
Retry ActiveStorage FileNotFoundError when processing variants

calebtocco:
https://swell.radicalbear.com/tasks/32666


app/controllers/rad_common/attachments_controller.rb
```diff
@@ -58,7 +58,7 @@ def set_variant
         klass = params[:class_name].classify.constantize
         record = klass.find_decoded(params[:id])
         begin
-          @variant = record.send(params[:variant]).processed
+          @variant = RadicalRetry.perform_request(retry_count: 2) { record.send(params[:variant]).processed }
```
calebtocco, line 61: Using same retry count used in the downloader

PR #267, opened 2020-11-06
Built in validator

elbzero:
https://swell.radicalbear.com/tasks/33106

### DIFF
--- a/app/services/rad_common/search.rb
+++ b/app/services/rad_common/search.rb
@@ -22,8 +22,8 @@ module RadCommon
       @params = params
       @search_name = search_name
       @filtering = Filtering.new(filters: filters, search: self)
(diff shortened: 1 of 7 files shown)


app/services/rad_common/search.rb
```diff
@@ -39,6 +39,10 @@ def valid?
       @filtering.validate_params
     end
 
+    def invalid?
```
garyfoster, line 42: Is there a need to have `invalid?` and `valid?` not be private? If not, we should make them private since the built-in validator handles this. I can take care of fixing any existing projects.

elbzero: yeah cause the way I automatically invoke this is by calling in the actual view. See _filter.html.haml it's a bit of an odd pattern to invoke the flash message here but was trying to figure out a way to do this without having to add extra boiler plate code. Though I could make valid? private since I only call invalid? in the view.

PR #285, opened 2021-01-18
Allow excluding ID's in autocomplete

elbzero:
Task 33170


app/services/global_autocomplete.rb
```diff
@@ -29,97 +29,98 @@ def global_super_search_result
...
-        column_values.push(format_column_value(record[column_name]))
+    def autocomplete_result(scope)
+      return [] unless scope && Pundit.policy!(user, GlobalAutocomplete.check_policy_klass(user, klass)).index?
+
+      self.current_scope = scope
+      order = scope[:query_order] || 'created_at DESC'
+      query = Pundit.policy_scope!(user, GlobalAutocomplete.check_policy_klass(user, klass))
+      query = query.joins(joins) if joins
+      query = query.where(where_query, search: "%#{params[:term]}%").order(order)
+      query = query.where.not(id: params[:excluded_ids]) if params[:excluded_ids].present?
```
garyfoster, line 50: this will fail if there are joins or includes in the query but will document that tech debt for later

garyfoster:
also, need to raise an exception if this is called with super search as that wouldn't make sense with multiple tables

PR #287, opened 2021-01-26
Equals filter

elbzero:
https://swell.radicalbear.com/tasks/33461


app/services/rad_common/equals_filter.rb
```diff
@@ -0,0 +1,47 @@
...
+  # This is used to generate an input used for a SQL like filter
+  class EqualsFilter
+    attr_reader :column
+    attr_reader :data_type
+
+    ##
+    # @param [String] column the database column that is being filtered
+    # @param [Symbol] type the type of filter
+    # @param [Symbol] data_type controls what the input type is based on data type of column
+    def initialize(column:, type:, data_type:)
```
garyfoster, line 12: @elbzero rubocop says the `type` arg is not used?

elbzero: @garyfoster this is because of how we build the search filters dynamically [here](https://github.com/radicalbear/rad_common/blob/68ddac4d07aa793f3335dabda4c17a5eb377153b/app/services/rad_common/filtering.rb#L53) this is how we do all of the filters. There might be a better way to do this but we are required to pass in all the params even if even if they aren't really useful to the final class like DateFilter. Hit me up on slack if you want to discuss further

garyfoster: can it be changed to `_type`? That is usually how you name an arg that isn't used, but not sure if that messes anything up in this case

elbzero: @garyfoster it can't because thats what we use to pass in stuff like type: RadCommon:DateFilter. However I figured out another way around, by deleting the type param [here](https://github.com/radicalbear/rad_common/commit/0be31496806f6e9a016497f7941c6b4a356c0234)

app/services/rad_audit_search.rb
```diff
@@ -30,7 +30,8 @@ def filters_def
        { input_label: 'Action',
          column: :action,
          options: %w[create update destroy] },
-       { column: :remote_address, type: RadCommon::LikeFilter }]
+       { column: :remote_address, type: RadCommon::LikeFilter },
+       { column: :auditable_id, type: RadCommon::EqualsFilter, data_type: :string }]
```
garyfoster, line 34: curious, this is an integer field but you're setting data type to string?

elbzero: you're right this should be integer. Forgot to change this back when I was testing

PR #385, opened 2022-01-31
Heroku Commands Check For Valid App

calebtocco:
Tasks 35948

### DIFF
--- a/app/services/heroku_commands.rb
+++ b/app/services/heroku_commands.rb
@@ -2,6 +2,7 @@ class HerokuCommands
   class << self
     def backup(app_name)
       check_production do
+        check_valid_app(app_name)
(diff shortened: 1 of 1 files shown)


app/services/heroku_commands.rb
```diff
@@ -164,5 +167,10 @@ def remove_accounting_keys
       def write_log(message)
         puts message
       end
+
+      def check_valid_app(app_name)
+        _output, error = Open3.capture3("heroku apps:info #{app_option(app_name)}")
```
calebtocco, line 172: this is just an easy and safe command used as a preflight check, nothing really special about `apps:info`

garyfoster: that is awesome dude!

PR #387, opened 2022-02-04
Allow New Autocomplete

calebtocco:
Tasks 35407

### DIFF
--- a/app/assets/javascripts/rad_common/autocomplete-rails.js
+++ b/app/assets/javascripts/rad_common/autocomplete-rails.js
@@ -126,16 +126,30 @@
           if(jQuery(e).data('excluded-ids')) {
               params['excluded_ids'] = jQuery(e).data('excluded-ids');
           }
+          const allowNewModel = jQuery(e).data('allow-new-model');
           jQuery.getJSON( jQuery(e).attr('data-autocomplete'), params, function() {
+            if (allowNewModel) {
+              arguments[0].unshift({ id: '', label: `Create "${request.term}" ${allowNewModel}` });
+            }
             if(arguments[0].length == 0) {
               arguments[0] = []
-              arguments[0][0] = { id: "", label: "no existing match" }
+              if (!allowNewModel) {
+                arguments[0][0] = { id: '', label: 'no existing match' };
+              }
             }
+            let newOptionExists = false;
             jQuery(arguments[0]).each(function(i, el) {
               var obj = {};
               obj[el.id] = el;
+              // Check if record already exists in suggestions
+              if (el.label === request.term) {
+                newOptionExists = true;
+              }
               jQuery(e).data(obj);
             });
+            if (allowNewModel && newOptionExists) {
+              arguments[0].shift(); // Remove create option if record exists in suggestions
+            }
             response.apply(null, arguments);
           });
         },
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/rad_common/autocomplete-rails.js
```diff
@@ -126,16 +126,30 @@
           if(jQuery(e).data('excluded-ids')) {
               params['excluded_ids'] = jQuery(e).data('excluded-ids');
           }
+          const allowNewModel = jQuery(e).attr('data-allow-new-model');
           jQuery.getJSON( jQuery(e).attr('data-autocomplete'), params, function() {
+            if (!!allowNewModel) {
+              arguments[0].unshift({ id: '', label: `Create "${request.term}" ${allowNewModel}` });
```
calebtocco, line 132: adds create text if allow new model, doesnt really do anything, just closes the dropdown

app/assets/javascripts/rad_common/autocomplete-rails.js
```diff
@@ -126,16 +126,30 @@
...
+            }
             if(arguments[0].length == 0) {
               arguments[0] = []
-              arguments[0][0] = { id: "", label: "no existing match" }
+              if (!allowNewModel) {
+                arguments[0][0] = { id: '', label: 'no existing match' };
```
calebtocco, line 137: not relevant, replaced by create option

app/assets/javascripts/rad_common/autocomplete-rails.js
```diff
@@ -126,16 +126,30 @@
           if(jQuery(e).data('excluded-ids')) {
               params['excluded_ids'] = jQuery(e).data('excluded-ids');
           }
+          const allowNewModel = jQuery(e).attr('data-allow-new-model');
```
elbzero, line 129: @calebtocco I think you can do
`jQuery(e).data('allow-new-model'`

which will automatically cast it to truthy value so no need for !!

app/views/layouts/_simple_autocomplete.html.haml
```diff
@@ -11,14 +11,15 @@
   tooltip = local_assigns[:tooltip].presence ||
             "Search for the intended #{association.to_s.humanize}, "\
             'then select it from the dropdown that appears below.'
-
+  input_name = local_assigns[:allow_new] ? "#{base_object}[#{association}_name]" : "#{association}_name_search"
```
elbzero, line 14: @calebtocco I wonder if we really need to rename it here. I kind of like it still being named search, its a little clearer what its doing and used for.

spec/dummy/app/models/division.rb
```diff
@@ -3,6 +3,9 @@ class Division < ApplicationRecord
   include Hashable
 
   belongs_to :owner, class_name: 'User'
+  belongs_to :category, optional: true
+
+  attr_accessor :category_name
```
elbzero, line 8: @calebtocco I wish there was a way we could do this without using an attr_accessor and dynamically add it or move it into a reusable concern on the controller or something like that, I prefer less wiring up if possible, but I think its good for now since that would probably be a lot more work maybe thats a phase 2 thing.

elbzero (commented):
Just a couple comments, overall looks good!

spec/dummy/app/models/division.rb
calebtocco: right, we should be able to improve on this in the future

app/views/layouts/_simple_autocomplete.html.haml
calebtocco: true, we can leave the search and change the accessor

app/assets/javascripts/rad_common/autocomplete-rails.js
calebtocco: cool ill give that a try

PR #391, opened 2022-02-06
Fix Array Field Schema Validations

calebtocco:
Tasks 36009

### DIFF
--- a/app/models/concerns/schema_validations.rb
+++ b/app/models/concerns/schema_validations.rb
@@ -37,6 +37,8 @@ module SchemaValidations
       column_type = column.type
       if respond_to?(:defined_enums) && defined_enums.has_key?(column.name)
         :enum
+      elsif column.array?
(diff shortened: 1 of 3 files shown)


app/models/concerns/schema_validations.rb
```diff
@@ -68,6 +70,8 @@ def validate_presence(column, name, datatype)
         validate_logged :validates_inclusion_of, name, in: [true, false], message: :blank
       elsif !column.default.nil? && column.default.blank?
         validate_logged :validates_with, NotNilValidator, attributes: [name]
+      elsif datatype == :array
+        validate_logged :validates, name, presence: true, if: -> { public_send(name).nil? }
```
calebtocco, line 74: kinda funky but we cant use exclusion here b/c of a change in rails 6.1 https://github.com/rails/rails/issues/41051, and this works fine

PR #394, opened 2022-02-12
Autocomplete Dropdown Toggle

calebtocco:
Tasks 36050


app/presenters/association_input_presenter.rb
```diff
@@ -0,0 +1,54 @@
...
+    def initial_value
+      local_assigns[:initial_value].presence || record.public_send(association)&.to_s
+    end
+
+    def record
+      form.object
+    end
+
+    def collection
+      Pundit.policy_scope(current_user, record.association(association).klass).sorted
```
garyfoster, line 52: we usually use `policy_scope!` to ensure the scope exists, i'm fixing

PR #411, opened 2022-03-09
Rad Retry Raise Original Flag

calebtocco:
Tasks 35918, 35791

### DIFF
--- a/app/services/phone_sms_sender.rb
+++ b/app/services/phone_sms_sender.rb
@@ -16,7 +16,7 @@ class PhoneSMSSender
   end
 
   def send!
-    RadicalRetry.perform_request do
+    RadicalRetry.perform_request(raise_original: true) do
       if mms?
         twilio.send_mms to: to_number, message: message, media_url: media_url
       else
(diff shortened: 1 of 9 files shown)


app/services/phone_sms_sender.rb
```diff
@@ -16,7 +16,7 @@ def initialize(message, from_user_id, to_mobile_phone, media_url, force_opt_out)
   end
 
   def send!
-    RadicalRetry.perform_request do
+    RadicalRetry.perform_request(raise_original: true) do
```
calebtocco, line 19: we retry the lookups twice, maybe we should limit this too? failures to send still cost $ im assuming

garyfoster: nah, it's ok - the lookups can get hit a lot more with entering phone numbers -  volume on sending SMS will be a fraction of that

garyfoster: nah, it's ok - the lookups can get hit a lot more with entering phone numbers -  volume on sending SMS will be a fraction of that

PR #423, opened 2022-03-29
Allow non-admin users to manage users

garyfoster:
Tasks 36471


app/controllers/users_controller.rb
```diff
@@ -54,15 +55,13 @@ def create
   end
 
   def update
-    if params[:user][:password].blank? && params[:user][:password_confirmation].blank?
```
garyfoster, line 57: moved this into a before action to fix hound complexity issue

app/policies/user_policy.rb
```diff
@@ -3,6 +3,19 @@ def index?
     user.internal?
   end
 
+  def create?
+    user.permission?(:manage_user)
+  end
+
+  def update_security_roles?
```
garyfoster, line 10: this new policy action controls the controller and the user form for this scenario

app/policies/user_policy.rb
```diff
@@ -3,6 +3,19 @@ def index?
     user.internal?
   end
 
+  def create?
+    user.permission?(:manage_user)
```
garyfoster, line 7: this policy allows non-admins to manage users

PR #441, opened 2022-04-29
move action styles to webpacker

elbzero:
https://swell.radicalbear.com/tasks/36806


garyfoster:
doing this on main branch, before we do Rails 7 - https://github.com/radicalbear/rad_common/pull/442

PR #473, opened 2022-08-15
Radical spec support

calebtocco:
37653
37654


lib/generators/rad_common/install/rails_helper.rb
```diff
@@ -38,7 +39,8 @@
 # directory. Alternatively, in the individual `*_spec.rb` files, manually
 # require only the support files necessary.
 #
-Dir[Rails.root.join('spec/support/**/*.rb')].sort.each { |f| require f }
```
garyfoster, line 41: will this fail on projects that have additional items in spec/support?

PR #475, opened 2022-08-18
Reactivate Users Button

calebtocco:
Tasks 37637


app/models/concerns/radbear_user.rb
```diff
@@ -191,6 +191,10 @@ def test_sms!(from_user)
     UserSMSSenderJob.perform_later 'Test SMS', from_user.id, id, nil, false
   end
 
+  def reactivate
+    update(last_activity_at: nil)
```
garyfoster, line 195: use `update!` so this can't silently fail

garyfoster: nevermind :)

PR #481, opened 2022-09-01
Onboarding

garyfoster:
Task 37598


app/helpers/rad_common/application_helper.rb
```diff
@@ -188,6 +201,18 @@ def export_button(model_name)
...
 
+    def show_help_menu?(user)
+      !user_signed_in? || user.external?
+    end
+
+    def onboarded?
+      Onboarding.new(current_user).onboarded?
+    end
+
+    def onboarding
```
garyfoster, line 212: @elbzero where it this used, can you link me to an example?

app/helpers/rad_common/application_helper.rb
```diff
@@ -188,6 +201,18 @@ def export_button(model_name)
               class: 'btn btn-secondary btn-sm')
     end
 
+    def show_help_menu?(user)
+      !user_signed_in? || user.external?
+    end
+
+    def onboarded?
```
garyfoster, line 208: do we need this? can this usage be moved inside the view partial [here](https://github.com/cannapages/cannasaver_admin/blob/main/app/views/layouts/_navigation.html.haml#L10)?

garyfoster: @elbzero  do we need this? can this usage be moved inside the view partial [here](https://github.com/cannapages/cannasaver_admin/blob/main/app/views/layouts/_navigation.html.haml#L10)?

PR #500, opened 2022-09-27
Getter method for the full name of a state given the abbreviation

jayywolff:
Simple getter method to grab the full name of a state given the abbreviation.

Usage: 
```ruby

StateOptions.full_name('FL') # returns 'Florida'

```


garyfoster:
thanks @jayywolff looks great

PR #511, opened 2022-10-10
Cleanup datetime picker

garyfoster:
Task 37777


app/inputs/datetime_local_input.rb
```diff
@@ -1,15 +0,0 @@
-class DatetimeLocalInput < SimpleForm::Inputs::StringInput
```
garyfoster, line 1: @elbzero is this still used?

PR #520, opened 2022-10-24
Saved Filters

calebtocco:
Task 38100


app/models/saved_search_filter.rb
```diff
@@ -0,0 +1,10 @@
+class SavedSearchFilter < ApplicationRecord
+  belongs_to :user
```
garyfoster, line 2: need the has_many on the rad user concern, probably should be dep destroy

PR #546, opened 2022-12-02
start on attachments layout

elbzero:
https://swell.radicalbear.com/tasks/40544


app/views/layouts/_attachment.html.haml
```diff
@@ -1,41 +1,3 @@
...
-      .attachment-button-box
-        = link_to attachment_label,
-                  url_for(attachment),
-                  target: '_blank',
-                  rel: :noopener,
-                  class: 'btn btn-secondary attachment-link text-wrap'
-
-        = delete_link
+  - attachment = record.send(attachment_name)
+  = render 'layouts/attachment_object', attachment: attachment, record: record
```
garyfoster, line 3: doesn't this need to pass all those optional variables to the next partial? maybe we should create a helper and eventually phase out the layouts

app/helpers/rad_common/attachments_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+      return unless record.persisted?
+
+      attachments = record.send(attachment_name)
+      return if attachments.blank?
+
+      record.send(attachment_name).map { |attachment|
+        render_attachment_object(attachment: attachment, record: record, override_label: override_label,
+                                 show_filename: show_filename, no_delete_button: no_delete_button,
+                                 override_path: override_path, new_tab: new_tab)
+      }.join(' ').html_safe
```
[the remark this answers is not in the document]
elbzero: @garyfoster is there another way we do this? I can't remember

app/helpers/rad_common/attachments_helper.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      attachment = record.send(attachment_name)
+      return unless !attachment.respond_to?(:attached?) || attachment.attached?
+
+      render_attachment_object(attachment: attachment, record: record, override_label: override_label,
+                               show_filename: show_filename, no_delete_button: no_delete_button,
+                               override_path: override_path, new_tab: new_tab)
+    end
+
+    def render_attachment_object(attachment:, record:,
```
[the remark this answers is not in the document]
elbzero: @garyfoster not sure if there is a way around this? maybe I just add an exception in rubocop.xml ?

app/helpers/rad_common/attachments_helper.rb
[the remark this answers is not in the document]
elbzero: actually nvm, I see we use safe_join

app/helpers/rad_common/attachments_helper.rb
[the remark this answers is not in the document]
garyfoster: ya, gotta add an override

app/views/layouts/_attachment.html.haml
```diff
@@ -1,41 +1 @@
...
-
-      .attachment-button-box
-        = link_to attachment_label,
-                  url_for(attachment),
-                  target: '_blank',
-                  rel: :noopener,
-                  class: 'btn btn-secondary attachment-link text-wrap'
-
-        = delete_link
+= render_attachment_layout(attachment_name: attachment_name, record: record)
```
garyfoster, line 1: don't we need to pass the rest of the variable options here so that existing uses of the layout won't fail?

app/views/layouts/_attachment_object.html.haml
```diff
@@ -0,0 +1,9 @@
+- if attachment
```
garyfoster, line 1: can this be done with a helper rather than a layout, just to keep it all in ruby since the rest is?

app/views/layouts/_attachments.html.haml
```diff
@@ -0,0 +1,2 @@
+- record.send(attachment_name).each do |attachment|
```
garyfoster, line 1: I assumed we would do a helper for this as well instead of a layout

elbzero: done

app/views/layouts/_attachment.html.haml
elbzero: maybe I should just nix the whole layout and then we can update it to just use the helper? Shouldn't be too bad just need to remember to do it for each project

PR #550, opened 2022-12-22
Global search performance enhancement

jayywolff:
Make global search check for record existence with `none?` rather than `count.zero?`

I saw a huge performance improvement on an app with a large db with this one simple modification, passing it along upstream.
This method appears to be called for every search scope in the global search feature which is rendered in the internal navbar, so it makes a huge impact on system performance during business hours and page load speeds

### DIFF
--- a/app/services/global_search.rb
+++ b/app/services/global_search.rb
@@ -40,6 +40,6 @@ class GlobalSearch
     end
 
     def no_records?(scope)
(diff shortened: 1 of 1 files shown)


garyfoster:
That is awesome @jayywolff, thank you! We've had a few of these count vs. none vs. size type of issues. They see to fall under the radar of scout but so they are harder to find, thank you again.

PR #553, opened 2022-12-26
User tweaks

garyfoster:
Task 40830


app/services/rad_common_routes.rb
```diff
@@ -37,6 +37,10 @@ def self.extended(router)
         resources :json_web_tokens, only: :new
       end
 
+      authenticate :user, ->(u) { u.external? } do
```
garyfoster, line 40: this is only used in one project (ABN) - it's still protected by UserPolicy as well

PR #578, opened 2023-03-08
Refactor devise-authy to devise-twilio-verify gem

garyfoster:
Task 41441


lib/rad_common/engine.rb
```diff
@@ -10,7 +10,6 @@ class Engine < ::Rails::Engine
     require 'authtrail'
     require 'csv'
     require 'devise'
-    require 'devise-authy'
```
jayywolff, line 13: if you haven't already, you need to replace this line with 
`require 'devise-twilio-verify'

PR #595, opened 2023-04-20
Reformat Twilio Log Numbers

calebtocco:
Task 41692


spec/dummy/db/migrate/20230420102508_update_twilio_log_number_format.rb
```diff
@@ -0,0 +1,24 @@
+class UpdateTwilioLogNumberFormat < ActiveRecord::Migration[6.1]
```
garyfoster, line 1: wow @calebtocco look at that SQL

PR #613, opened 2023-05-31
Rad Config for Devise Twilio Remember Device

calebtocco:
Task 42123

### DIFF
--- a/app/services/rad_config.rb
+++ b/app/services/rad_config.rb
@@ -298,6 +298,10 @@ class RadConfig
       array_config_item! :restricted_audit_attributes
     end
 
+    def twilio_verify_remember_device!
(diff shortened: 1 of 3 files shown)


app/services/rad_config.rb
```diff
@@ -148,6 +148,18 @@ def twilio_verify_service_sid!
...
+    def twilio_verify_enabled?
+      boolean_config_item! :twilio_verify_enabled
+    end
+
+    def twilio_verify_all_users?
+      boolean_config_item! :twilio_verify_all_users
+    end
+
+    def twilio_verify_remember_device!
+      config_item! :twilio_verify_remember_device_days
```
garyfoster, line 160: we currently have a precedent for requiring all config settings in all apps so that it's explicit - I could go either way on it but for now, following pattern

PR #630, opened 2023-07-18
Duplicate notification

elbzero:
Task 37856

### DIFF
--- a/app/models/concerns/duplicate_fixable.rb
+++ b/app/models/concerns/duplicate_fixable.rb
@@ -209,7 +209,9 @@ module DuplicateFixable
     end
 
     def maybe_notify_duplicates
-      return if duplicate.score.blank?
+      process_duplicates
+
+      return if duplicate&.score.nil?
 
       Notifications::PossibleDuplicateFoundNotification.main.notify!(self)
     end
--- a/app/models/concerns/duplicate_fixable.rb
+++ b/app/models/concerns/duplicate_fixable.rb
@@ -211,7 +211,7 @@ module DuplicateFixable
     def maybe_notify_duplicates
       process_duplicates
 
-      return if duplicate&.score.nil?
+      return if reload.duplicate&.score.nil?
 
       Notifications::PossibleDuplicateFoundNotification.main.notify!(self)
     end
--- a/app/models/concerns/duplicate_fixable.rb
+++ b/app/models/concerns/duplicate_fixable.rb
@@ -17,6 +17,8 @@ module DuplicateFixable
     scope :high_duplicates, lambda {
       joins(:duplicate).where('score IS NOT NULL AND score >= ? AND sort <= 500', score_upper_threshold)
     }
+
+    after_create :maybe_notify_duplicates
   end
 
   module ClassMethods
@@ -204,6 +206,12 @@ module DuplicateFixable
 
     def model_klass
       self.class.to_s.constantize
+    end
+
+    def maybe_notify_duplicates
+      return if duplicate.score.blank?
+
+      Notifications::PossibleDuplicateFoundNotification.main.notify!(self)
     end
 
     def table_name
--- a/app/models/concerns/duplicate_fixable.rb
+++ b/app/models/concerns/duplicate_fixable.rb
@@ -18,7 +18,7 @@ module DuplicateFixable
       joins(:duplicate).where('score IS NOT NULL AND score >= ? AND sort <= 500', score_upper_threshold)
     }
 
-    after_create :maybe_notify_duplicates
+    after_commit :maybe_notify_duplicates, on: :create
   end
 
   module ClassMethods
--- a/app/models/notifications/possible_duplicate_found_notification.rb
+++ b/app/models/notifications/possible_duplicate_found_notification.rb
@@ -1,14 +1,21 @@
 module Notifications
   class PossibleDuplicateFoundNotification < ::NotificationType
     def absolute_user_ids
-      created_by_user = subject_record.audits.order(created_at: :desc).first&.user
       return [created_by_user.id] if created_by_user.present?
 
       []
     end
 
-    def subject_record
-      nil
+    def created_by_user
+      subject_record.audits.order(created_at: :desc).first&.user
+    end
+
+    def mailer_message
+      "Possible duplicate (#{subject_record}) entered by #{created_by_user}"
+    end
+
+    def description
+      'Possible duplicate found'
     end
   end
 end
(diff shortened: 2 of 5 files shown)


app/models/concerns/duplicate_fixable.rb
```diff
@@ -17,6 +17,8 @@ module DuplicateFixable
     scope :high_duplicates, lambda {
       joins(:duplicate).where('score IS NOT NULL AND score >= ? AND sort <= 500', score_upper_threshold)
     }
+
+    after_commit :maybe_notify_duplicates, on: :create
```
garyfoster, line 21: I assumed this would be on the `Duplicate` model?

app/models/concerns/duplicate_fixable.rb
```diff
@@ -206,6 +208,13 @@ def model_klass
       self.class.to_s.constantize
     end
 
+    def maybe_notify_duplicates
+      process_duplicates
```
garyfoster, line 212: this shouldn't be needed?

spec/dummy/db/migrate/20230718181431_add_duplicate_notification.rb
```diff
@@ -0,0 +1,7 @@
+class AddDuplicateNotification < ActiveRecord::Migration[7.0]
+  def change
+    return unless Rails.env.production? && SecurityRole.count.positive?
+
+    Notifications::PossibleDuplicateFoundNotification.create! security_roles: SecurityRole.all
```
garyfoster, line 5: these are auto-created now on demand, no need for migration

app/models/concerns/duplicate_fixable.rb
elbzero: updated this

app/models/concerns/duplicate_fixable.rb
elbzero: updated this

spec/dummy/db/migrate/20230718181431_add_duplicate_notification.rb
elbzero: updated this

app/models/duplicate.rb
```diff
@@ -1,3 +1,11 @@
 class Duplicate < ApplicationRecord
   belongs_to :duplicatable, polymorphic: true
+
+  after_commit :maybe_notify_duplicates, on: :create
+
+  def maybe_notify_duplicates
```
garyfoster, line 6: not sure if this is handled but we only want to notify when it's over the threshold, same logic that drives the red badge in the top nav and whether or not the fix duplicates button shows up

app/models/notifications/possible_duplicate_found_notification.rb
```diff
@@ -0,0 +1,25 @@
+module Notifications
+  class PossibleDuplicateFoundNotification < ::NotificationType
+    def absolute_user_ids
+      return [created_by_user.id] if created_by_user.present?
```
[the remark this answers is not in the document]
garyfoster: 1. seems fine, that's how I would do it - a tad bit fragile but a better way would probably be overkill - we should have auditing turned on for all models that can be duplicatable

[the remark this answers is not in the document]
garyfoster: 2. yes, how about a separate notification to the admins rather than cc'ing them, less shame

app/models/notifications/possible_duplicate_found_notification.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      []
```
garyfoster, line 6: i'm thinking raise an exception in this case so we can investigate why this would be nil - wondering if we might hit a race condition if this is triggered before audit is created, any way to find out?

app/models/notifications/possible_duplicate_found_notification.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def created_by_user
```
garyfoster, line 9: seems like some of these methods should be private?

app/models/notifications/possible_duplicate_found_notification.rb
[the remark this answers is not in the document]
elbzero: sounds good, added an extra notification

app/models/notifications/possible_duplicate_found_notification.rb
elbzero: agreed

app/models/duplicate.rb
elbzero: I'm calling the "duplicates" method it by default only returns those over the lower threshold see here https://github.com/radicalbear/rad_common/blob/4b64b03f1e9aa06513931d850eda7cd405e3490e/app/models/concerns/duplicate_fixable.rb#L130

garyfoster: nah, we want the same logic that decides whether or not the badge shows up in the nav, for example see [here](https://github.com/radicalbear/rad_common/blob/main/app/helpers/duplicates_helper.rb#L23)

app/models/notifications/possible_duplicate_admin_notification.rb
```diff
@@ -0,0 +1,17 @@
...
+    end
+
+    def description
+      'Possible duplicate found'
+    end
+
+    private
+
+      def created_by_user
+        subject_record.audits.where(action: 'create').first&.user
```
garyfoster, line 14: suggest we raise an exception if this is ever blank, and do you think there could be a race condition where this is triggered before the audit is created?

app/models/notifications/possible_duplicate_found_notification.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def description
+      'Possible duplicate found'
+    end
+
+    private
+
+      def created_by_user
+        subject_record.audits.where(action: 'create').first&.user
```
garyfoster, line 24: suggest we raise an exception if this is ever blank, and do you think there could be a race condition where this is triggered before the audit is created?

app/models/duplicate.rb
elbzero: Ok gotcha, I will update that, out of curiosity, whats the use case for the lower threshold, I see it used here 
https://github.com/radicalbear/rad_common/blob/5b132b6f139ddd320b7a4f76521ced8e386742c3/app/presenters/card_presenter.rb#L227

I was just wondering if that logic is either redundant and/or possibly conflicting?

app/models/notifications/possible_duplicate_admin_notification.rb
elbzero: done, see other comment for info about race condition

app/models/duplicate.rb
garyfoster: I don't see the lower threshold used there, am I missing it?

elbzero: @garyfoster I'm not sure anymore, maybe I was reading something wrong? I think I'm good now

PR #634, opened 2023-08-10
Refactor Wicked PDF framework into rad_common

elbzero:
Task 42271


app/services/rad_config.rb
```diff
@@ -202,6 +202,10 @@ def storage_config_override?
       boolean_config_item! :storage_config_override
     end
 
+    def use_wicked?
```
garyfoster, line 205: remove per discussion

lib/generators/rad_common/install/install_generator.rb
```diff
@@ -48,6 +48,11 @@ def create_initializer_file
         copy_file '../../../../../spec/dummy/config/initializers/simple_form_components.rb',
                   'config/initializers/simple_form_components.rb'
 
+        if RadConfig.use_wicked?
```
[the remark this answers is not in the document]
elbzero: yep removing

lib/generators/rad_common/install/install_generator.rb
```diff
@@ -48,6 +48,9 @@ def create_initializer_file
         copy_file '../../../../../spec/dummy/config/initializers/simple_form_components.rb',
                   'config/initializers/simple_form_components.rb'
 
+        copy_file '../../../../../spec/dummy/config/initializers/wicked_pdf.rb',
```
garyfoster, line 51: seems like we won't need this file in the projects, when I remove it from better_way, the PDF's still work, I think it can use the initializer in rad_common

elbzero: @garyfoster which initializer in rad_common?

spec/dummy/config/initializers/wicked_pdf.rb
```diff
@@ -0,0 +1,22 @@
+# WickedPDF Global Configuration
```
garyfoster, line 1: I think this may need to be moved to config/initializers so that it is active in the other projects

lib/generators/rad_common/install/install_generator.rb
garyfoster: see the below comment, I think it needs to be moved

elbzero: its removed

spec/dummy/config/initializers/wicked_pdf.rb
elbzero: its moved

spec/dummy/app/javascript/src/app_specific_pdf.scss
```diff
@@ -0,0 +1 @@
+/* Custom PDF styles here */
```
garyfoster, line 1: do we still need this?

elbzero: removing

PR #644, opened 2023-10-08
Style some things based on active/inactive

garyfoster:
Task 42816


app/helpers/rad_common/application_helper.rb
```diff
@@ -149,6 +149,12 @@ def bootstrap_flash_close_button
       tag.button(sanitize('&times;'), type: 'button', class: 'close', 'data-dismiss': 'alert')
     end
 
+    def table_row_style(record, style_class: 'table-danger')
+      return unless record.present? && record.respond_to?(:active?) && !record.active?
```
garyfoster, line 153: should we remove the `record.respond_to?(:active?)` and just require that we always implement `active?` on every model so we never forget any of these?

app/presenters/card_presenter.rb
```diff
@@ -128,6 +128,17 @@ def output_icon
     raise 'missing card header icon'
   end
 
+  def card_style
+    unless %w[show edit].include?(action_name) &&
+           instance.present? &&
+           instance.respond_to?(:active?) &&
```
garyfoster, line 134: should we remove the `instance.respond_to?(:active?)` and just require that we always implement `active?` on every model so we never forget any of these?

app/views/login_activities/index.html.haml
```diff
@@ -8,14 +8,14 @@
...
-      - @login_activities.each do |login_activity|
-        %tr{ class: login_activity.success ? '' : 'alert-danger' }
-          %td= format_datetime(login_activity.created_at)
-          %td= login_activity.identity
-          %td= format_boolean(login_activity.success)
-          %td= login_activity.failure_reason
-          %td= login_activity.ip
-          %td= login_activity.user_agent
-          %td= login_activity.referrer
+      - @login_activities.each do |record|
```
garyfoster, line 11: i've wanted to do this for a while, to change the model name to `record` so that the boilerplate is more resilient - we don't have to do this on every index page though but just did these while I was in there

spec/dummy/lib/templates/haml/scaffold/index.html.haml
```diff
@@ -14,21 +14,21 @@
     - if show_actions
           %th.actions Actions
 
-      - @<%= plural_table_name %>.each do |<%= singular_table_name %>|
+      - @<%= plural_table_name %>.each do |record|
```
garyfoster, line 17: renamed this to `record` so the boilerplate is more resilient when copying/pasting etc

PR #665, opened 2023-11-15
Seeder Refactor and Accessible State

calebtocco:


app/models/application_record.rb
```diff
@@ -4,6 +4,9 @@ class ApplicationRecord < ActiveRecord::Base
   class_attribute :schema_validations_loaded
```
garyfoster, line 4: this file is copied via the installer but I like this better, just need to remove it from the installer

PR #678, opened 2023-12-05
Changes to pending user status

garyfoster:
Task 43518

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -17,7 +17,7 @@ module Users
           flash[:success] = "We invited '#{resource}' and sent them an email. You can update their settings as " \
                             'needed here.'
 
-          redirect_to edit_user_path(resource)
+          redirect_to users_path
         else
           flash[:error] = "Could not add user: #{user.errors.full_messages.join(', ')}"
         end
--- a/app/mailers/notification_mailer.rb
+++ b/app/mailers/notification_mailer.rb
@@ -2,17 +2,17 @@ class NotificationMailer < RadMailer
   before_action :enable_settings_link
 
   def new_user_signed_up(recipients, user)
-    auto_approve = user.auto_approve?
+    user_is_active = user.user_status.active?
 
     action_message = 'Review their user registration information'
-    action_message += auto_approve ? ' if desired.' : ' and approve them if desired.'
+    action_message += user_is_active ? ' if desired.' : ' and approve them if desired.'
 
     @email_action = { message: action_message,
-                      button_text: auto_approve ? 'Review' : 'Review & Approve',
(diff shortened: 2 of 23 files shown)


app/controllers/users/invitations_controller.rb
```diff
@@ -17,7 +17,7 @@ def create
           flash[:success] = "We invited '#{resource}' and sent them an email. You can update their settings as " \
                             'needed here.'
 
-          redirect_to edit_user_path(resource)
```
garyfoster, line 20: When inviting a user we no longer will need to edit their status so we're going to return to the index page, and the newly added user will be on the top of the list in the sort.

app/mailers/notification_mailer.rb
```diff
@@ -2,17 +2,17 @@ class NotificationMailer < RadMailer
   before_action :enable_settings_link
 
   def new_user_signed_up(recipients, user)
-    auto_approve = user.auto_approve?
```
garyfoster, line 5: the auto approve switch is no longer used and removed

app/models/concerns/rad_user.rb
```diff
@@ -80,8 +80,7 @@ module RadUser
     after_initialize :twilio_verify_default, if: :new_record?
     before_validation :check_defaults
     before_validation :set_timezone, on: :create
-    after_save :notify_user_approved
```
garyfoster, line 83: just tightening this up with best practices

app/models/concerns/rad_user.rb
```diff
@@ -133,11 +132,6 @@ def admin?
     permission?(:admin)
   end
 
-  def auto_approve?
```
garyfoster, line 136: this feature is no longer used and just getting it out of the way to avoid confusion, we can easily add it back or something like it if needed

app/models/concerns/rad_user.rb
```diff
@@ -286,10 +286,11 @@ def set_timezone
     end
 
     def notify_user_approved
-      return if auto_approve?
+      return unless RadConfig.pending_users?
+      return if invited_to_sign_up?
 
-      return unless saved_change_to_user_status_id? && user_status &&
```
garyfoster, line 291: just tightening this up

spec/dummy/db/migrate/20231205185433_pending_user_status.rb
```diff
@@ -0,0 +1,8 @@
+class PendingUserStatus < ActiveRecord::Migration[7.0]
+  def change
+    return if UserStatus.none?
+    return if RadConfig.pending_users?
+
+    execute "DELETE FROM user_statuses WHERE name = 'Pending';"
```
garyfoster, line 6: I'll deal with this manually on each site, I don't think there will be any existing pending users, but if there are, this will blow up

PR #686, opened 2023-12-29
Misc updates

garyfoster:


app/views/login_activities/index.html.haml
```diff
@@ -8,14 +8,14 @@
       %tr
         = render 'layouts/sort_headers', search: @login_activity_search
 
-      - @login_activities.each do |login_activity|
+      - @login_activities.each do |record|
```
garyfoster, line 11: rename this variable to record so it's easier when copying code and stuff

spec/dummy/app/controllers/attorneys_controller.rb
```diff
@@ -66,6 +66,6 @@ def set_attorney
 
     def permitted_params
       params.require(:attorney).permit(:first_name, :last_name, :middle_name, :company_name, :phone_number, :email,
-                                       :address_1, :address_2, :city, :state, :zipcode)
+                                       :address_1, :address_2, :city, :state, :zipcode, :active)
```
garyfoster, line 69: adding this in preparation for another update down the road that needs this

PR #688, opened 2023-12-29
Marketing website features

garyfoster:
Task 10

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,6 +4,11 @@ class PagesController < ApplicationController
   def home
     skip_authorization
 
+    if marketing_site?
+      @marketing_site = true
+      return
+    end
+
     if user_signed_in?
       redirect_to RadConfig.start_route!
     else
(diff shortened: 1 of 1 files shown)


app/controllers/pages_controller.rb
```diff
@@ -23,4 +26,10 @@ def terms
   def privacy
     skip_authorization
   end
+
+  private
+
+    def marketing_site?
+      Rails.env.production? && request.host.start_with?('www.')
```
garyfoster, line 33: switching things by environment here is bad practice, maybe use a flag in rad_common.yml instead

app/views/layouts/_navigation_right.html.haml
```diff
@@ -2,6 +2,8 @@
   - if user_signed_in?
     - unread_count = current_user.notifications.unread.count
 
+    - unless Rails.env.test?
```
garyfoster, line 5: switching things by environment here is bad practice, maybe use a flag in rad_common.yml instead

PR #693, opened 2024-01-01
Rad rate limiter eric

elbzero:
Task 10

### DIFF
--- a/spec/dummy/config/environments/test.rb
+++ b/spec/dummy/config/environments/test.rb
@@ -25,7 +25,7 @@ Rails.application.configure do
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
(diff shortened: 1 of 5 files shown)


spec/dummy/config/environments/test.rb
```diff
@@ -25,7 +25,7 @@
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
-  config.cache_store = :null_store
+  config.cache_store = :memory_store
```
garyfoster, line 28: the installer will overwrite this, should/can we make this standard or will that mess something up in other projects?

elbzero: @garyfoster we could make it standard, I don't see any reason why not, and if we run into an issue we can back track as needed, @calebtocco what do you think?

PR #750, opened 2024-02-21
Fix sed errors in installer

garyfoster:
Task 2947

### DIFF
--- a/.github/workflows/rad_update_bot.yml
+++ b/.github/workflows/rad_update_bot.yml
@@ -19,6 +19,7 @@ jobs:
       POSTGRES_DB: rad_common_development
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
+      CI: true
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/rad_update_bot.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       POSTGRES_DB: rad_common_development
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
+      CI: true
```
garyfoster, line 22: using the same pattern and seems like a good name for this environment variable since the bot is arguably part of our continuous integration machinery

lib/generators/rad_common/install/install_generator.rb
```diff
@@ -271,12 +271,20 @@ def apply_migration(source)
...
+          search_and_replace_type search, replace, 'haml'
+          search_and_replace_type search, replace, 'rake'
           return unless js
 
-          system "find . -type f -name \"*.js\" -print0 | xargs -0 sed -i '' -e 's/#{search}/#{replace}/g'"
+          search_and_replace_type search, replace, 'js'
+        end
+
+        def search_and_replace_type(search, replace, file_type)
+          if ENV['CI']
```
garyfoster, line 283: this is a difference in sed syntax between MacOS and Linux

PR #778, opened 2024-04-12
[3090] Contact Logs

calebtocco:
Task 3090


app/mailers/rad_mailer.rb
```diff
@@ -114,4 +115,43 @@ def escape_name(recipient_name)
     def enable_settings_link
       @notification_settings_link = true
     end
+
+    def create_contact_log
+      ContactLog.create!(
+        from_email: mail.from.first,
+        reply_to: mail.reply_to.first,
+        subject: mail.subject,
+        message: @message, # What to store here if anything? could be html template
```
garyfoster, line 124: maybe just a limited excerpt? let's discuss

app/mailers/rad_mailer.rb
```diff
@@ -114,4 +115,43 @@ def escape_name(recipient_name)
...
+        log_type: :outgoing,
+        service_type: :sendgrid
+      ).tap do |log|
+        create_contact_log_attachments(log) # Do we want to store all of these?
```
garyfoster, line 128: probably not, seems like a lot - why do we have this attachments model to begin with?

app/models/contact_log_recipient.rb
```diff
@@ -0,0 +1,70 @@
+class ContactLogRecipient < ApplicationRecord
+  belongs_to :contact_log
+  belongs_to :to_user, class_name: 'User', optional: true
+
+  enum email_type: { to: 0, cc: 1, bcc: 2 }
+  # TODO: Determine twilio and sendgrid status differences
```
garyfoster, line 6: let's discuss

app/models/contact_log_recipient.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  validates_with PhoneNumberValidator, fields: [{ field: :phone_number }], skip_twilio: true
+  validates_with EmailAddressValidator, fields: [:email], skip_sendgrid: true
+
+  validate :validate_service_type
+  validate :validate_incoming_fields
+
+  before_validation :check_success
+
+  # TODO: Verify that this is method is actually used
```
garyfoster, line 35: agreed

app/services/rad_common/search.rb
```diff
@@ -192,5 +192,11 @@ def active_users
       def inactive_users
         Pundit.policy_scope!(current_user, User).inactive.by_name
       end
+
+      def enum_scopes(model, enum)
```
garyfoster, line 196: this is awesome, probably should have a maint update to refactor existing projects

app/mailers/rad_mailer.rb
garyfoster: let's ask @elbzero if he knows, maybe we can get rid of it but if we need to keep it, and we don't store the attachments on emails but do store them on SMS then we might want to rename the table/model so it's not confusing

app/mailers/rad_mailer.rb
garyfoster: we decided to replace subject and message with one field named content

app/models/contact_log_recipient.rb
garyfoster: I couldn't find any use for it, removed

PR #781, opened 2024-04-15
User reverse sort option

garyfoster:
Task 3412


app/controllers/security_roles_controller.rb
```diff
@@ -3,7 +3,7 @@ class SecurityRolesController < ApplicationController
 
   def index
     authorize SecurityRole
-    @security_roles = policy_scope(SecurityRole.by_name).page(params[:page])
```
garyfoster, line 6: renaming this to be more standardized, not related to this task though

PR #802, opened 2024-06-02
Refactor email (SendGrid) contact logs

garyfoster:
Task 3910


app/mailers/concerns/contact_mailer.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  included do
+    before_action :start_contact_log
+    after_action :finish_contact_log
+  end
+
+  private
+
+    def start_contact_log
+      # TODO: is there a better way to handle this content required field?
```
garyfoster, line 12: I tried playing around with different attempts at `before_action`, `around_action` and `after_action` but gave up without digging in too much and just did this workaround. I don't love adding a dummy value for content and then updating the other attributes later but it does seem to work. Perhaps there is a better way.

app/mailers/concerns/contact_mailer.rb
```diff
@@ -0,0 +1,35 @@
...
+      end
+
+      return if mail.bcc.blank?
+
+      mail.bcc.each { |recipient| ContactLogRecipient.create!(contact_log: log, email: recipient, email_type: :bcc) }
+    end
+
+    def rad_headers
+      headers['X-SMTPAPI'] = { unique_args: { host_name: RadConfig.host_name!,
+                                              contact_log_id: @contact_log_id } }.to_json
```
garyfoster, line 33: it's a little bit of a chicken/egg problem, we need the `contact_log_id` to send through to SendGrid but can't get that ID until we create the contact log, which needs some mail attributes to fully populate the records

app/mailers/concerns/contact_mailer.rb
```diff
@@ -0,0 +1,35 @@
+module ContactMailer
```
garyfoster, line 1: concerns

app/mailers/concerns/contact_mailer.rb
```diff
@@ -0,0 +1,35 @@
...
+  extend ActiveSupport::Concern
```
garyfoster, line 2: concerns work with mailers, woooop!

app/models/contact_log.rb
```diff
@@ -19,6 +19,9 @@ class ContactLog < ApplicationRecord
 
   validates_with PhoneNumberValidator, fields: [{ field: :from_number }], skip_twilio: true
 
+  audited
```
garyfoster, line 22: figure this could be helpful one day

app/models/contact_log_recipient.rb
```diff
@@ -37,6 +37,9 @@ class ContactLogRecipient < ApplicationRecord
 
   before_validation :check_success
 
+  audited
```
garyfoster, line 40: figure this could be helpful one day

app/models/contact_log.rb
```diff
@@ -34,7 +37,7 @@ def self.opt_out_message_sent?(to_number)
   private
 
     def validate_incoming
-      errors.add(:sent, 'must be true') unless sms_sent?
+      errors.add(:sms_sent, 'must be true') unless sms_sent?
```
garyfoster, line 40: fixed an existing bug

app/models/login_activity.rb
```diff
@@ -3,7 +3,7 @@ class LoginActivity < ApplicationRecord
 
   scope :recent_first, -> { order(id: :desc) }
   scope :successful, -> { where(success: true) }
-  scope :failure, -> { where(success: false) }
```
garyfoster, line 6: just changed terminology here to be consistent

app/services/rad_sendgrid_status_receiver.rb
```diff
@@ -64,6 +65,16 @@ def process_status!
       end
     end
 
+    def update_contact_log!
+      return unless suppression?
+
+      contact_log.contact_log_recipients.find_by!(email: email).update! email_status: event
```
garyfoster, line 71: this will hard fail if 1) the contact log can't be found and 2) the contact log recipient can't be found with the email address of the suppression event from SendGrid - thinking we let it roll and see how it goes once in the wild - in theory it should never fail

app/services/login_activity_search.rb
```diff
@@ -36,14 +36,18 @@ def filters_def
...
+       { label: 'Successful' },
+       { label: 'Failed' },
        { label: 'IP', column: 'ip' },
        { label: 'Agent', column: 'user_agent' },
        { column: 'referrer' }]
     end
 
     def failure_reasons
-      Pundit.policy_scope!(current_user, LoginActivity).failure.select(:failure_reason).distinct.map(&:failure_reason)
+      Pundit.policy_scope!(current_user, LoginActivity)
```
garyfoster, line 47: fixed an existing sniff here

app/services/rad_sendgrid_status_receiver.rb
garyfoster: should we  check to see if there are multiple recipients with the same email address and update them all, or perhaps prevent email from being duplicated within a contact log's contact log recipients?

app/mailers/concerns/contact_mailer.rb
garyfoster: duh, done

spec/services/rad_rate_limiter_spec.rb
```diff
@@ -22,7 +22,7 @@
       before { 4.times { rate_limiter.run { 'foo' } } }
 
       it 'raises error' do
-        expect { rate_limiter.run { 'foo' } }.not_to raise_error(described_class::RateLimitExceededError)
```
garyfoster, line 25: totally unrelated, but fixed this annoying warning that showed up every test suite run

app/services/rad_sendgrid_status_receiver.rb
garyfoster: I added a raise so it will blow up if it happens and we can investigate, also added a related spec

PR #819, opened 2024-06-25
fix email subject for failed contacts

garyfoster:
Task 4436


app/models/contact_log_recipient.rb
```diff
@@ -72,7 +72,6 @@ def validate_incoming_fields
       return if contact_log.blank? || contact_log.outgoing? || contact_log.email?
 
       errors.add(:to_user_id, 'must be blank') if to_user_id.present?
-      errors.add(:sms_status, 'must be blank') if sms_status.present?
```
garyfoster, line 75: this is a duplicate validation, its already defined above

PR #824, opened 2024-07-06
Rad 4504 show route exists

garyfoster:
Task 4504

### DIFF
--- a/app/helpers/rad_common/application_helper.rb
+++ b/app/helpers/rad_common/application_helper.rb
@@ -13,7 +13,10 @@ module RadCommon
     end
 
     def show_route_exists_for?(record)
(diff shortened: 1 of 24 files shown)


app/helpers/rad_common/application_helper.rb
```diff
@@ -13,7 +13,10 @@ def secured_link(resource, format: nil)
     end
 
     def show_route_exists_for?(record)
-      Rails.application.routes.url_helpers.respond_to? "#{record.class.table_name.singularize}_path"
```
garyfoster, line 16: this was not taking into consideration the :get method so that a route that had update, destroy, etc but no show route would still return true

PR #829, opened 2024-07-10
Preparing for a shakeup to how active/inactive users work

garyfoster:
Task 4524

### DIFF
--- a/app/models/concerns/rad_user.rb
+++ b/app/models/concerns/rad_user.rb
@@ -33,7 +33,7 @@ module RadUser
 
     attr_accessor :approved_by, :do_not_notify_approved, :initial_security_role_id
 
-    scope :active, -> { joins(:user_status).where('user_statuses.active = TRUE') }
+    scope :active, -> { joins(:user_status).where(user_statuses: { active: true }) }
     scope :admins, -> { active.by_permission 'admin' }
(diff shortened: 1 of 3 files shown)


app/models/concerns/rad_user.rb
```diff
@@ -33,7 +33,7 @@ module RadUser
 
     attr_accessor :approved_by, :do_not_notify_approved, :initial_security_role_id
 
-    scope :active, -> { joins(:user_status).where('user_statuses.active = TRUE') }
```
garyfoster, line 36: just improved syntax, no changes

app/models/concerns/rad_user.rb
```diff
@@ -100,12 +100,8 @@ def to_s
     end
   end
 
-  def active
```
garyfoster, line 103: removed the `active` method, we've been using `active?` instead going forward and just want to avoid confusion and have a method that is easier to search for when the time comes to refactor anything

app/models/concerns/rad_user.rb
```diff
@@ -167,7 +163,7 @@ def other_audits_created
   end
 
   def display_style
-    if user_status.active || (RadConfig.pending_users? && user_status == UserStatus.default_pending_status)
+    if user_status.active? || (RadConfig.pending_users? && user_status == UserStatus.default_pending_status)
```
garyfoster, line 166: just improved syntax

PR #842, opened 2024-07-22
require two factor authentication for all admin users

garyfoster:
Task 499


app/models/concerns/duplicate_fixable.rb
```diff
@@ -287,6 +287,7 @@ def fix_duplicate(key, user)
         duplicate_record.contact_logs_to.update_all to_user_id: id
         ContactLog.where(record_type: 'User', record_id: duplicate_record.id).update_all record_id: id
         Audited::Audit.where(user_id: duplicate_record.id).update_all user_id: id
+        User.where(invited_by_id: duplicate_record.id).update_all invited_by_id: id
```
garyfoster, line 290: unrelated change I found when merging dupes on a client's app

PR #848, opened 2024-08-01
Set From User via Options in RadMailer#simple_message

calebtocco:
Task 4853


app/views/layouts/rad_mailer.html.haml
```diff
@@ -60,7 +60,7 @@
                 .footer-item
                   Click
                   = link_to 'here', "#{protocol}://#{RadConfig.host_name!}/rad_common/notification_settings"
-                  to opt out or change your notification settings.
```
garyfoster, line 63: I worried that users might think this opts them out upon clicking, but it doesn't, it just takes them to the notification settings page

PR #851, opened 2024-08-05
separate user active from user status and authentication status

garyfoster:
Task 4524

### DIFF
--- a/app/models/concerns/rad_user.rb
+++ b/app/models/concerns/rad_user.rb
@@ -60,11 +60,7 @@ module RadUser
             "WHERE security_roles.#{permission} = TRUE)")
     }
 
-    scope :inactive, lambda {
-      joins(:user_status)
-        .where('user_statuses.active = FALSE OR (invitation_sent_at IS NOT NULL AND invitation_accepted_at IS NULL)')
-    }
-
+    scope :inactive, -> { joins(:user_status).where(user_statuses: { active: false }) }
     scope :not_inactive, -> { where.not(user_status_id: UserStatus.default_inactive_status.id) }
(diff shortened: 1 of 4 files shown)


app/models/concerns/rad_user.rb
```diff
@@ -60,11 +60,7 @@ module RadUser
             "WHERE security_roles.#{permission} = TRUE)")
     }
 
-    scope :inactive, lambda {
-      joins(:user_status)
-        .where('user_statuses.active = FALSE OR (invitation_sent_at IS NOT NULL AND invitation_accepted_at IS NULL)')
```
garyfoster, line 65: i'm not really sure what the ramifications are of this but I don't think we should combine the user status active logic with the devise authentication active logic

app/models/concerns/rad_user.rb
```diff
@@ -74,7 +70,9 @@ module RadUser
     validate :validate_twilio_verify
     validate :validate_mobile_phone
     validate :password_excludes_name
-    validates :security_roles, presence: true, if: :active?
+
+    # this should be changed to "if: :active?" at some point, see Task 2024
+    validates :security_roles, presence: true, if: :active_for_authentication?
```
garyfoster, line 75: had to change this due to an existing "bug" that allows inviting a user w/o a role, we had an old task for that

app/models/concerns/rad_user.rb
```diff
@@ -101,7 +99,7 @@ def to_s
   end
 
   def active?
-    active_for_authentication?
+    user_status&.active?
```
garyfoster, line 102: this is the main logic change in this PR

PR #865, opened 2024-08-22
fix bug with SMS false positives

garyfoster:
Task 5067


app/models/contact_log_recipient.rb
```diff
@@ -62,6 +62,10 @@ def active?
     success?
   end
 
+  def to_s
```
garyfoster, line 65: unrelated change

PR #876, opened 2024-09-17
Don't notify on failure of email changed emails

garyfoster:
Task 5365


app/mailers/concerns/rad_contact_mailer.rb
```diff
@@ -20,31 +20,30 @@ def finish_contact_log
...
+        add_rad_recipient recipient, :bcc
       end
     end
 
+    def add_rad_recipient(recipient, type)
+      ContactLogRecipient.create! contact_log: @rad_contact_log,
+                                  email: recipient,
+                                  email_type: type,
+                                  email_status: :delivered,
+                                  notify_on_fail: @rad_notify_on_fail.nil? ? true : @rad_notify_on_fail
```
garyfoster, line 44: this is the key change on this file

app/mailers/rad_devise_mailer.rb
```diff
@@ -79,6 +79,8 @@ def invitation_instructions(record, token, opts = {})
   def email_changed(record, opts = {})
     @contact_log_record = record
     @contact_log_from_user = record
+    @rad_notify_on_fail = false
```
garyfoster, line 82: if a user has a bad email and when they change emails and it fails sending to the old one, we don't want to notify anyone

PR #895, opened 2024-10-15
Add decoding to email body

brockdenson:
[View the Task](https://app.radstack.com/tasks/5724)


spec/services/sendgrid_status_receiver_spec.rb
```diff
@@ -29,7 +29,7 @@
 
     expect(last_email.subject).to include 'Outgoing Email Failed'
     expect(last_email.body.encoded).to include 'Attorney'
-    expect(last_email.body.encoded).to include contact_log.content
+    expect(last_email.html_part&.decoded).to include contact_log.content
```
garyfoster, line 32: why safe nav here?

brockdenson: After reviewing, it looks like safe nav is unnecessary since `html_part` will always be present. Safe nav would prevent an error if `html_part` were `nil`. Went ahead and removed it

PR #905, opened 2024-10-23
allow audits page for a single item to be searchable

garyfoster:
Task 5884


app/controllers/audits_controller.rb
```diff
@@ -1,25 +1,12 @@
 class AuditsController < ApplicationController
   def index
```
garyfoster, line 2: I could use a review of the authorization for this endpoint

PR #912, opened 2024-11-05
enhance outgoing contact notifications for SMS contacts

garyfoster:
Task 5298


app/models/notifications/outgoing_contact_failed_notification.rb
```diff
@@ -5,15 +5,16 @@ def auth_mode
...
-              []
-            end
-
-      ids.delete(to_user.id) if to_user.present?
-      ids = SecurityRole.admin_role.users.active.pluck(:id) if ids.blank?
-      ids.delete(to_user.id) if to_user.present?
+      ids = init_ids
+
+      if contact_log.sms?
+        ids.push(to_user.id) if to_user.present?
```
garyfoster, line 11: basically, if it's an SMS that fails, just email the affected from/to user but not the admins

PR #927, opened 2024-11-17
streamline user notifications

garyfoster:
Task 6277


app/mailers/rad_mailer.rb
```diff
@@ -167,4 +168,10 @@ def validate_options(options, keys)
 
       raise "unknown options: #{unknown_keys}"
     end
+
+    def get_started_link
+      return RadConfig.config_item!(:get_started_link) if RadConfig.config_item(:get_started_link).present?
```
garyfoster, line 173: ABN's get started link goes to the mobile app site, not app.abn.live

app/models/concerns/rad_user.rb
```diff
@@ -247,6 +247,12 @@ def developer?
     email.end_with? RadConfig.developer_domain!
   end
 
+  class_methods do
+    def user_approved_message
```
garyfoster, line 251: this is shared in a couple places

app/models/concerns/rad_user.rb
```diff
@@ -365,7 +371,12 @@ def notify_user_approved
     end
 
     def notify_user_approved_user
-      RadMailer.your_account_approved(self).deliver_later
+      raise 'missing approved_by' if approved_by.blank?
```
garyfoster, line 374: I think this is only applicable to sign up apps like ABN, which had overridden behavior that also sent a text message to new users when they are approved. Made it standard here which doesn't seem like it will hurt anything.

PR #928, opened 2024-11-17
fix Heroku app name detection in bin scripts

garyfoster:
Task 6282

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,7 +6,6 @@ AllCops:
     - 'db/migrate/*'
     - 'lib/templates/**/*'
     - 'node_modules/**/*'
-    - 'bin/**/*'
     - 'config/puma.rb'
     - 'config/initializers/simple_form.rb'
     - 'config/initializers/simple_form_bootstrap.rb'
--- a/bin/rad_bin.rb
+++ b/bin/rad_bin.rb
@@ -1,3 +1,4 @@
+require 'English'
 require 'fileutils'
 
 module RadBin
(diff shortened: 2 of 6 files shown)


.rubocop.yml
```diff
@@ -6,7 +6,6 @@ AllCops:
     - 'db/migrate/*'
     - 'lib/templates/**/*'
     - 'node_modules/**/*'
-    - 'bin/**/*'
```
garyfoster, line 9: not 100% sure about this but I do link the linter to help me see code optimization tips, even in these bin files

bin/rad_bin.rb
```diff
@@ -1,3 +1,4 @@
+require 'English'
```
garyfoster, line 1: rubocop added this when I ran it with the auto correct flag (-A) - not sure why, but will research

bin/rad_bin.rb
```diff
@@ -14,18 +15,50 @@ def self.rad_common_not_permitted!
 
   def self.heroku_app_exists?(app_name)
     `heroku apps:info --app #{app_name}`
-    $?.success?
+    $CHILD_STATUS.success?
```
garyfoster, line 18: rubocop fix

bin/rad_bin.rb
garyfoster: seems fine, see [here](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/SpecialGlobalVars)

PR #945, opened 2024-12-11
Add validation for external security roles

brockdenson:
[View the Task](https://app.radstack.com/tasks/6248)

### DIFF
--- a/app/models/concerns/rad_security_role.rb
+++ b/app/models/concerns/rad_security_role.rb
@@ -64,8 +64,8 @@ module RadSecurityRole
     end
 
     def validate_external_security_role
-      if external && !RadConfig.external_users?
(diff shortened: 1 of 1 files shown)


app/models/concerns/rad_security_role.rb
```diff
@@ -61,4 +62,10 @@ def validate_rules
       errors.add(:allow_invite, 'is not applicable') if allow_invite? && RadConfig.disable_invite?
       errors.add(:allow_sign_up, 'is not applicable') if allow_sign_up? && RadConfig.disable_sign_up?
     end
+
+    def validate_external_security_role
+      return unless external && !RadConfig.external_users?
```
garyfoster, line 67: we like using the question mark version of the field name for all boolean fields, like `external?`

PR #950, opened 2024-12-13
clear domain name

garyfoster:
Task 4920


app/models/concerns/rad_user.rb
```diff
@@ -245,8 +243,8 @@ def timeout_in
     external? ? Devise.timeout_in : RadConfig.timeout_hours!.hours
   end
 
-  def rad_developer?
-    email.end_with? RAD_DOMAIN
+  def developer?
+    email.end_with? RadConfig.developer_domain!
```
garyfoster, line 247: i'll fix all apps by adding the config item

PR #961, opened 2025-01-02
policy problem with resolve duplicates

garyfoster:
Task 6882

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -50,10 +50,13 @@ class ApplicationPolicy
     user.permission?(:admin)
   end
 
+  def resolve_duplicates?
+    index?
+  end
+
   alias not_duplicate? destroy?
   alias reset_duplicates? destroy?
   alias merge_duplicates? destroy?
-  alias resolve_duplicates? index?
   alias duplicate_do_later? destroy?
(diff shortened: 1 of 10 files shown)


app/presenters/card_presenter.rb
```diff
@@ -242,7 +240,8 @@ def include_duplicate_action?
       action_name == 'show' &&
         RadCommon::AppInfo.new.duplicates_enabled?(klass.name) &&
         instance.duplicate.present? &&
-        instance.duplicate.score.present?
+        instance.duplicate.score.present? &&
+        instance_policy.resolve_duplicates?
```
garyfoster, line 244: this line is the only material change in this file, the rest is just refactoring to share code

app/policies/application_policy.rb
```diff
@@ -50,10 +50,13 @@ def report?
...
   end
 
+  def resolve_duplicates?
+    index?
+  end
+
   alias not_duplicate? destroy?
   alias reset_duplicates? destroy?
   alias merge_duplicates? destroy?
-  alias resolve_duplicates? index?
```
garyfoster, line 56: the alias resolves to the index method in this file and not the subclass

PR #966, opened 2025-01-03
Fix policy issues with global search and searchable associations

garyfoster:
Task 557

### DIFF
--- a/app/services/global_autocomplete.rb
+++ b/app/services/global_autocomplete.rb
@@ -1,13 +1,15 @@
 class GlobalAutocomplete
   include RadCommon::ApplicationHelper
 
-  attr_reader :params, :search_scopes, :user
+  attr_reader :params, :search_scopes, :user, :mode
   attr_accessor :current_scope
 
-  def initialize(params, search_scopes, user)
+  def initialize(params, search_scopes, user, mode)
     @params = params
     @search_scopes = search_scopes
     @current_scope = selected_scope
+    @mode = mode
+    validate_mode
     validate_global_search_scope
     @user = user
   end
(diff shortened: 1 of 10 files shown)


app/services/global_autocomplete.rb
```diff
@@ -166,4 +174,15 @@ def column_def(column)
     def scope_with_where?(scope)
       (scope[:columns].present? && scope[:columns].any?) || scope[:query_where].present?
     end
+
+    def policy_ok?
```
garyfoster, line 178: here's the key change, if it's a global search, it needs both index and show, but if it's a searchable association, it just needs index

app/services/global_search.rb
```diff
@@ -35,11 +34,28 @@ def scopes
...
+
     def hide_scope?(scope)
       scope[:hide_global_nav] || no_records?(scope)
     end
 
     def no_records?(scope)
       Pundit.policy_scope!(current_user, scope[:model].constantize).none?
     end
+
+    def policy_ok?(item)
```
garyfoster, line 51: here's the key change, if it's a global search, it needs both index and show, but if it's a searchable association, it just needs index

app/views/layouts/_global_search.html.haml
```diff
@@ -48,3 +50,4 @@
               %input.global_search_id{ type: 'hidden', name: 'global_search_id' }
               %input.global_search_model_name{ type: 'hidden', name: 'global_search_model_name' }
               %input.global_search_scope{ type: 'hidden', name: 'global_search_scope' }
+              %input.global_search_mode{ type: 'hidden', name: 'global_search_mode', value: 'global_search' }
```
garyfoster, line 53: not sure if there is a better way to do this but it seems to work

PR #970, opened 2025-01-04
Fix policy issues with global search and searchable associations

garyfoster:


app/controllers/search_controller.rb
```diff
@@ -37,4 +42,16 @@ def global_search_result
       end
     end
   end
+
+  private
+
+    def set_global_search_mode
+      @global_search_mode = if params[:global_search_mode].blank?
+                              :global_search
```
garyfoster, line 50: instead of passing this from the global search uses, I'll just fall back to it when blank

PR #975, opened 2025-01-10
tweak rubo rules

garyfoster:
Task 4697

### DIFF
--- a/.shared_rubocop.yml
+++ b/.shared_rubocop.yml
@@ -21,6 +21,7 @@ AllCops:
     - 'config/initializers/new_framework_defaults*'
     - 'spec/spec_helper.rb'
     - 'config.ru'
+    - 'bin/*'
 
 # Want to use even indentation for the various methods within a class, and
 # also want to be able to indent private methods without errors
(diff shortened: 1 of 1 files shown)


.shared_rubocop.yml
```diff
@@ -21,6 +21,7 @@ AllCops:
     - 'config/initializers/new_framework_defaults*'
     - 'spec/spec_helper.rb'
     - 'config.ru'
+    - 'bin/*'
```
garyfoster, line 24: these are generated in the other projects but the rubocop rules will still run in rad_common because the AllCops is overridden in .rubocop.yml in the rad_common project

.shared_rubocop.yml
```diff
@@ -50,6 +51,11 @@ Lint/RaiseException:
 Lint/StructNewOverride:
   Enabled: true
 
+Lint/RedundantDirGlobSort:
+  Enabled: true
+  Exclude:
+    - 'spec/rails_helper.rb'
```
garyfoster, line 57: this has some boilerplate that we don't want to change even though it fails rubo

.shared_rubocop.yml
```diff
@@ -114,6 +120,11 @@ Rails/UnknownEnv:
 Rails/Output:
   Enabled: true
 
+Rails/RootPathnameMethods:
+  Enabled: true
+  Exclude:
+    - 'spec/rails_helper.rb'
```
garyfoster, line 126: this has some boilerplate that we don't want to change even though it fails rubo

PR #984, opened 2025-01-17
copy perm attach updates

elbzero:
Task 2974


elbzero:
FYI this is a special branch on rad_common branched from rad_common from the same point in history as rad_common that cannasaver_public main is using. This is because cannasaver_public is broken with the new ESBuild requirements so we did this as temp hack.

In theory we can close this branch altogether once cannasaver_public is updated to main, but should double check functionality still works when switching

PR #995, opened 2025-02-05
gracefully handle audits when model was removed from the project

garyfoster:
Task 7592


app/views/audits/index.html.haml
```diff
@@ -18,7 +18,7 @@
           - if audit.action == 'destroy'
             %td #{audit.auditable_type} (#{audit.auditable_id})
           - else
-            %td= audit_model_link(audit, audit.auditable)
+            %td= audit_model_link(audit, safe_auditable(audit))
```
garyfoster, line 21: Here's what this will look like when the model is missing:

Screenshot 2025-02-05 at 8 25 02 AM

PR #999, opened 2025-02-08
Allow admin to view other user notifications

garyfoster:
Task 7644

### DIFF
--- a/app/models/concerns/rad_user.rb
+++ b/app/models/concerns/rad_user.rb
@@ -42,6 +42,7 @@ module RadUser
     scope :recent_first, -> { order('users.created_at DESC') }
     scope :recent_last, -> { order('users.created_at') }
     scope :except_user, ->(user) { where.not(id: user.id) }
+    scope :with_notifications, -> { where('users.id IN (SELECT DISTINCT user_id FROM notifications)') }
 
     scope :sorted, lambda {
       if RadConfig.last_first_user?
--- a/app/policies/contact_log_recipient_policy.rb
+++ b/app/policies/contact_log_recipient_policy.rb
@@ -1,5 +1,2 @@
 class ContactLogRecipientPolicy < ContactLogPolicy
-  def index?
-    false
-  end
 end
--- a/app/policies/notification_policy.rb
+++ b/app/policies/notification_policy.rb
@@ -21,6 +21,8 @@ class NotificationPolicy < ApplicationPolicy
 
   class Scope < Scope
     def resolve
+      return scope.all if user.admin?
+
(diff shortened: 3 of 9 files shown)


app/models/concerns/rad_user.rb
```diff
@@ -42,6 +42,7 @@ module RadUser
     scope :recent_first, -> { order('users.created_at DESC') }
     scope :recent_last, -> { order('users.created_at') }
     scope :except_user, ->(user) { where.not(id: user.id) }
+    scope :with_notifications, -> { where('users.id IN (SELECT DISTINCT user_id FROM notifications)') }
```
garyfoster, line 45: this is used to limit the user filter list on the notifications search to show only users that have any notifications

app/policies/contact_log_recipient_policy.rb
```diff
@@ -1,5 +1,2 @@
 class ContactLogRecipientPolicy < ContactLogPolicy
-  def index?
-    false
```
garyfoster, line 3: this was causing the contact log system usage items to not show up, I handle this by hiding the index button in the UI instead

app/policies/notification_policy.rb
```diff
@@ -21,6 +21,8 @@ def destroy?
 
   class Scope < Scope
     def resolve
+      return scope.all if user.admin?
```
garyfoster, line 24: show all notifications if user is an admin

app/services/notification_search.rb
```diff
@@ -12,21 +12,35 @@ def initialize(params, current_user)
...
     def filters_def
-      [{ start_input_label: 'Start Date',
-         end_input_label: 'End Date',
-         column: :created_at,
-         type: RadCommon::DateFilter },
-       { column: 'notification_type_id', options: notification_type_options },
-       { column: 'content', type: RadCommon::LikeFilter },
-       { input_label: 'Record Type', column: 'record_type', options: record_type_options },
-       { input_label: 'Record ID', column: 'record_id', type: RadCommon::EqualsFilter, data_type: :integer }]
+      items = []
```
garyfoster, line 15: shows/hides the user filter and sort if user is an admin, but defaults the filter to current user

app/services/system_usage_search.rb
```diff
@@ -64,7 +64,12 @@ def date_column_ranges
   end
 
   def usage_items
-    @usage_items ||=
+    @usage_items ||= prepare_usage_items.sort_by { |item| item.is_a?(String) ? item : item.first }
```
garyfoster, line 67: sorts the system usage items in the table of the UI

PR #1000, opened 2025-02-10
ActionText attachments and Duplicates Weight Tweaks

elbzero:
Task 7658
Task 5341


app/services/duplicates_processor.rb
```diff
@@ -234,7 +234,7 @@ def calc_string_weight(record, duplicate_record, field_name, weight)
 
     def calc_email_weight(record, duplicate_record, weight)
       if email_compare?(record, duplicate_record)
-        weight
+        weight * 2 # TODO: make this an option to double the weight for key contact info
```
garyfoster, line 237: i'm including these 2 weight adjustments in this unrelated PR to save time but they should be moved into their own PR when time permits

PR #1004, opened 2025-02-18
ActionText attachments

elbzero:
Task 7777


app/controllers/valid_attachment_types_controller.rb
```diff
@@ -0,0 +1,8 @@
+class ValidAttachmentTypesController < ApplicationController
+  def index
+    skip_policy_scope
+    skip_authorization
+
+    render json: [RadCommon::VALID_ATTACHMENT_TYPES], status: :ok
```
garyfoster, line 6: maybe expand this endpoint to also include file size limit of 100mb, which is going to be shared by the template helper that checks

app/controllers/valid_attachment_types_controller.rb
```diff
@@ -0,0 +1,21 @@
+class ValidAttachmentTypesController < ApplicationController
```
garyfoster, line 1: this should probably be renamed now that we're changing the way it's used from the original

app/controllers/valid_attachment_types_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  def index
+    skip_policy_scope
+    skip_authorization
+
+    errors = []
+    unless RadCommon::VALID_ATTACHMENT_TYPES.include?(params[:content_type])
+      errors << "Invalid attachment type #{params[:content_type]}"
+    end
+
+    errors << 'Invalid attachment, must be 100 MB or less' if bytes_to_megabytes(params[:size].to_i) > 100
```
garyfoster, line 11: fyi, you can do things like `100.megabytes`, a built in method

app/controllers/valid_attachment_types_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    render json: { success: errors.empty?, errors: errors }, status: :ok
+  end
+
+  private
+
+    def bytes_to_megabytes(bytes)
```
garyfoster, line 18: no need for this, see above

app/helpers/rad_common/application_helper.rb
```diff
@@ -259,6 +259,14 @@ def rad_toast_data(template_locals)
         controller: 'toast' }
     end
 
+    def check_blob_validity(blob)
```
garyfoster, line 262: refactor this with the controller, it's doing the same thing and the logic should be shared

app/views/active_storage/blobs/_blob.html.erb
```diff
@@ -0,0 +1,16 @@
+<% check_blob_validity blob %>
```
garyfoster, line 1: convert to haml

app/helpers/rad_common/application_helper.rb
garyfoster: we should also add a spec to /shared if possible that ensures this raise happens on each project

PR #1007, opened 2025-02-27
prompt user if their timezone changes to see if they want to change it on the site

garyfoster:
Task 7902


app/services/user_timezone.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def ignore!
+    user.update! ignored_timezone: user.detected_timezone
+  end
+
+  private
+
+    def detected_timezone(ip_address)
+      Rails.cache.fetch("ip_address_time_zone:#{ip_address}", expires_in: 1.hour) do
+        raw_zone = Geocoder.search(ip_address).first.data['timezone']
```
garyfoster, line 35: this returns nil for local ip addresses, but might need to handle this if it happens on others

Gemfile.lock
```diff
@@ -23,6 +23,7 @@ PATH
       devise_invitable
       factory_bot_rails
       faker
+      geocoder
```
garyfoster, line 26: we have a few projects that use geocoder that can be refactored now that rad_common will have the gem

app/services/user_timezone.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  private
+
+    def local_ip_address?(ip_address)
+      %w[127.0.0.1 ::1].include?(ip_address)
+    end
+
+    def detected_timezone(ip_address)
+      Rails.cache.fetch("ip_address_time_zone:#{ip_address}", expires_in: 1.week) do
```
garyfoster, line 38: the 1 week expiry seems fine here since I doubt an IP address's location will even change at all

PR #1015, opened 2025-03-06
Add timezone detection specs

brockdenson:
Task 7902

### DIFF
--- a/spec/services/user_timezone_spec.rb
+++ b/spec/services/user_timezone_spec.rb
@@ -8,11 +8,7 @@ RSpec.describe UserTimezone, type: :service do
   let(:user_ip) { '192.168.1.1' }
   let(:new_ip) { '203.0.113.45' }
 
-  before do
-    allow_any_instance_of(described_class)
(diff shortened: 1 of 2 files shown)


spec/services/user_timezone_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+RSpec.describe UserTimezone, type: :service do
+  let(:user) { create :user, timezone: existing_timezone, detected_timezone: detected_timezone }
+  let(:existing_timezone) { 'Eastern Time (US & Canada)' }
+  let(:detected_timezone) { 'Pacific Time (US & Canada)' }
+  let(:new_detected_timezone) { 'Mountain Time (US & Canada)' }
+  let(:user_ip) { '192.168.1.1' }
+  let(:new_ip) { '203.0.113.45' }
+
+  before do
```
garyfoster, line 11: I prefer single line when possible and to use {} instead of do

garyfoster: fix other instances of this please

PR #1022, opened 2025-03-10
hack around ES Build for React apps

garyfoster:
Task 6761


Gemfile
```diff
@@ -27,5 +27,7 @@ end
 
 group :development, :test do
   gem 'bootsnap', require: false
+  gem 'jsbundling-rails'
```
garyfoster, line 30: we need to install these 2 gems to each project's Gemfile via the installer

.github/workflows/rspec_tests_react.yml
```diff
@@ -0,0 +1,124 @@
+name: RSpec Tests
```
garyfoster, line 1: this file will be installed for react apps

PR #1026, opened 2025-03-13
re-record VCR cassettes

garyfoster:
Task 693


spec/shared/models/concerns/contactable_spec.rb
```diff
@@ -199,10 +199,11 @@
       let(:address_2) { nil }
       let(:city) { 'Mordor' }
       let(:state) { 'FL' }
-      let(:zipcode) { '32816' }
```
garyfoster, line 202: this was showing as valid which is weird, but Smarty support helped me pick a zip that is invalid

PR #1039, opened 2025-03-27
clone local updates

elbzero:
Task 8479


rad_common.gemspec
```diff
@@ -82,4 +82,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'vcr'
   s.add_development_dependency 'webmock'
   s.add_development_dependency 'wkhtmltopdf-binary'
+  s.add_development_dependency 'tty-prompt'
```
elbzero, line 85: @calebtocco when you get to this, do you remember how this works, if I'm adding a dev dependency does that mean that we will need to add it to the Gem file for each project?

lib/rad_common/engine.rb
```diff
@@ -40,5 +40,6 @@ class Engine < ::Rails::Engine
     require 'turbo-rails'
     require 'twilio-ruby'
     require 'wicked_pdf'
+    require 'tty-prompt'
```
garyfoster, line 43: if it's only for development, I think this can be moved into the Gemfile in rad_common

PR #1042, opened 2025-03-27
Rad 8509 configurable company logo

elbzero:
Task 8509


app/controllers/companies_controller.rb
```diff
@@ -1,5 +1,6 @@
 class CompaniesController < ApplicationController
   before_action :set_company, only: %i[show edit update]
+  include ActiveStorageDownloader
```
garyfoster, line 3: curious whey this is needed? I kinda forget what it does

app/models/concerns/rad_company.rb
```diff
@@ -15,6 +15,8 @@ module RadCompany
 
     before_validation :clean_domain_spaces
 
+    has_one_attached :app_logo
```
garyfoster, line 18: we also need favicon, is that going to even be possible?

app/views/layouts/_navigation.html.haml
```diff
@@ -3,7 +3,7 @@
 .navbar.navbar-expand-lg.navbar-light.bg-light#navbar
   %div{ class: @marketing_site ? 'container' : 'container-fluid' }
     = link_to root_path, { class: 'navbar-brand' } do
-      = image_tag RadConfig.app_logo_filename!
```
garyfoster, line 6: assuming we'll remove that method from RadConfig

app/controllers/companies_controller.rb
elbzero: removing that, it was something I tried initially, but no longer needed

app/services/rad_config.rb
```diff
@@ -276,10 +276,6 @@ def favicon_filename!
       override_variable(:favicon_filename) || 'favicon.ico'
```
garyfoster, line 276: remove this method too?

elbzero: done

PR #1052, opened 2025-04-03
start on support for dynamic searchable input for search framework

elbzero:


app/services/rad_common/search_filter.rb
```diff
@@ -83,6 +85,12 @@ def filter_view
       'select'
     end
 
+    def searchable_scope?
+      return true if options.nil?
+
+      search_scope.present? && max_dropdown_size_exceeded?(options)
```
elbzero, line 91: This is support to dynamically use the search scope when max size is exceeded

app/services/user_grouper.rb
```diff
@@ -12,8 +13,20 @@ def call
     [me_item, internal_user_item, client_user_item, inactive_user_item].compact
   end
 
+  def user_filter(label, column, scope = nil)
+    default_options = { input_label: label, column: column, include_blank: false, blank_value_label: 'All Users' }
+    default_options[:scope] = scope if scope
+    return searchable_options(default_options) if max_dropdown_size_exceeded?(base_users)
```
elbzero, line 19: I'm manually changing the filter type here, right now I don't believe the search scope stuff works with grouped dropdowns, so in this case I would prefer the grouped dropdown if possible, but fallback to searchable dropdown if max exceeded.

I think we should be able to figure out how to support this in future, but just wanted to get us one step closer for now

app/inputs/searchable_association_input.rb
```diff
@@ -16,23 +16,15 @@ def input(wrapper_options = nil)
...
+        input_html_options.merge!(searchable_scope_options(show_subtext: options[:show_subtext],
+                                                           search_scope: options[:search_scope],
+                                                           excluded_ids: options[:excluded_ids]))
       else
         input_html_options[:class].push(:selectpicker)
       end
       options[:include_blank] = 'None' if options[:include_blank].nil?
     end
 
-    def search_options
```
elbzero, line 26: removing this stuff and now sharing the code between this simple form input and the search dropdown

app/helpers/rad_common/searchable_dropdown_helper.rb
```diff
@@ -0,0 +1,22 @@
+module RadCommon
+  MAX_DROPDOWN_SIZE = 300
+  module SearchableDropdownHelper
+    def searchable_scope_options(search_scope:, show_subtext: false, excluded_ids: nil)
```
elbzero, line 4: I moved this to a module to share this code, though not in love with this pattern if you have a better idea

PR #1072, opened 2025-04-21
Rad 8759 remove legacy rails config

brockdenson:
Task 8759

### DIFF
--- a/lib/generators/rad_common/install/install_generator.rb
+++ b/lib/generators/rad_common/install/install_generator.rb
@@ -492,7 +492,7 @@ gem 'rubocop-capybara'
   gem 'tty-prompt'
         RUBY
           end
-
+          
           unless RadConfig.legacy_assets?
             inject_into_file 'Gemfile', after: "gem 'bootsnap', require: false\n" do <<-'RUBY'
 gem 'jsbundling-rails'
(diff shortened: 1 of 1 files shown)


lib/generators/rad_common/install/install_generator.rb
```diff
@@ -342,6 +343,17 @@ def rad_config_setting_exists?(setting_name)
           File.readlines(RAD_CONFIG_FILE).grep(/#{setting_name}:/).any?
         end
 
+        def remove_legacy_rails_config_setting
+          return unless File.exist?(RAD_CONFIG_FILE)
```
garyfoster, line 347: this should always exist so it shouldn't be needed, we'd probably want it to crash if it's not there since something weird would be happening

lib/generators/rad_common/install/install_generator.rb
```diff
@@ -342,6 +343,17 @@ def rad_config_setting_exists?(setting_name)
...
+
+          content = File.read(RAD_CONFIG_FILE)
+
+          if content.match?(/^\s*legacy_rails_config:\s*.*\n/)
```
garyfoster, line 351: how about use the `rad_config_setting_exists?` method instead of this?

PR #1080, opened 2025-04-25
enable auditing for rich text (ActionText) fields

garyfoster:
Task 6965

### DIFF
--- a/app/helpers/rad_common/audits_helper.rb
+++ b/app/helpers/rad_common/audits_helper.rb
@@ -44,11 +44,7 @@ module RadCommon
     end
 
     def audit_model_link(audit, record)
-      label = if record.present? && record.respond_to?(:to_s)
-                "#{record.class} - #{record}"
-              else
-                "#{audit.auditable_type} (#{audit.auditable_id})"
-              end
+      label = audit_link_label(audit, record)
 
       return label if audit.nil? && record.nil?
       return link_to(label, record) if record.present? && show_route_exists_for?(record) && policy(record).show?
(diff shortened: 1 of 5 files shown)


spec/models/attorney_rich_text_auditing_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'rails_helper'
+
+RSpec.describe Attorney do
+  describe 'auditing rich text fields' do
+    subject(:update_notes) { attorney.update!(notes: 'Rich text content') }
```
garyfoster, line 5: although I'm not sure if this is wrong, we don't typically use a let variable for this situation, I would move the `attorney.update!` code into the {} below instead of extracting as a let

spec/models/attorney_rich_text_auditing_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    let(:attorney) { create :attorney }
+
+    it 'creates audits for the attorney and the rich text' do
+      expect { update_notes }.to change(Audited::Audit, :count).by(2)
+
+      rich_text_audit = Audited::Audit.where(auditable_type: 'ActionText::RichText').last
```
garyfoster, line 12: this can be a let variable though, please make that change

spec/models/attorney_rich_text_auditing_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+      expect(rich_text_audit.auditable.record).to eq(attorney)
```
garyfoster, line 14: the spec works great, good job

PR #1086, opened 2025-05-01
Rad 8881 clone local interactive mode

brockdenson:
Task 8881

### DIFF
--- a/app/services/heroku_commands.rb
+++ b/app/services/heroku_commands.rb
@@ -24,27 +24,20 @@ class HerokuCommands
       start_time = Time.now.utc
 
       write_log 'Dropping existing database schema'
-      write_log `psql -c "DROP SCHEMA public CASCADE; CREATE SCHEMA public" -h #{local_host} -U #{local_user} -d #{dbname}`
+      command = [
+        'psql -c',
+        '"DROP SCHEMA public CASCADE; CREATE SCHEMA public"',
+        "-h #{local_host}",
+        "-U #{local_user}",
+        "-d #{dbname}"
+      ].join(' ')
+      write_log `#{command}`
 
       restore_list_file = 'restore_list.txt'
-
-      case profile.to_s
-      when 'full'
-        write_log 'Generating full restore list (include all tables)'
-        write_log `pg_restore -l #{file_name} > #{restore_list_file}`
-      when 'exclude_audits'
-        write_log 'Generating restore list excluding audits table'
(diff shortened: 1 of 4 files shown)


lib/tasks/heroku.rake
```diff
@@ -3,11 +3,11 @@ namespace :heroku do
     HerokuCommands.backup args[:heroku_app]
   end
 
-  task :clone_local, %i[heroku_app backup_id] => :environment do |_t, args|
+  task :clone_local, %i[heroku_app profile backup_id] => :environment do |_t, args|
     # to run this with multiple args, the brackets must be escaped like this:
     # rails "heroku:clone_local[better-way-ars,a1010]"
```
elbzero, line 8: Can you update this comment to include some examples of what this command looks like now with the new profile command? This is a good place for reference for future

bin/clone_local
```diff
@@ -19,15 +21,54 @@ OptionParser.new { |opts|
 }.parse!
 
 FileUtils.chdir Dir.pwd do
-  heroku_app_name = RadBin.get_heroku_app_name(:all)
+  prompt = TTY::Prompt.new
+
+  if options[:interactive]
```
elbzero, line 26: This all seems to be good, though there does seem to be a lot going on in this one file, not sure if there is a way we can either move some of it into RadBin or even just methods within this file. 

Maybe separate out the different functionality, like interactive, remote, local, etc. No really "right" way to do it, but files like this can start to get pretty big over time

elbzero (changes requested):
Overall seems to work very good! This is awesome! 

Some comments about code refactoring, that would be good to do, to make sure stay clean over time

app/services/heroku_commands.rb
```diff
@@ -18,27 +20,38 @@ def backup(app_name)
...
-
-      write_log 'Changing Active Storage service to local'
-      ActiveStorage::Blob.update_all service_name: 'local'
-
       write_log 'Clearing certain production data'
       remove_user_avatars
       remove_encrypted_secrets
-      User.update_all twilio_verify_enabled: false
-      nil
+      reset_sensitive_local_data
```
elbzero, line 51: Moving to a method is fine, but I think we might need to preserve the original order in which this was called, trying bin/clone_local -i and restore from backup removing audits. Its giving me a weird error
Screenshot 2025-05-06 at 1 35 27 PM

brockdenson: Added the line `User.update_all twilio_verify_enabled: false` back to restore the original order

PR #1147, opened 2025-06-06
Fullcalendar updates

vincentdinic:
Task 9513


app/services/rad_nav/admin_menu.rb
```diff
@@ -27,6 +27,7 @@ def standard_items
          sidekiq,
          DropdownMenuItem.new(view_context, 'Company Info', view_context.company_edit_path),
          generate_jwt,
+         calendar,
```
garyfoster, line 30: make a "Divisions" menu that has 2 dropdown items, Divisions which goes to the index page and Calendar that goes to this page

garyfoster:
@vincentdinic the conflicts need fixing as well

PR #1150, opened 2025-06-26
add JS timezone detection in addition to IP address

garyfoster:
Task 8158


spec/dummy/app/javascript/application.js
```diff
@@ -9,3 +9,19 @@ RadCommon.setup();
 
 import './app_specific.js';
 import './controllers';
+
+document.addEventListener('DOMContentLoaded', () => {
```
garyfoster, line 13: this doesn't belong here, just using it to test for now

PR #1158, opened 2025-07-08
add validate user domains switch

elbzero:


spec/rails_helper.rb
```diff
@@ -133,22 +133,6 @@
 
   SpecSupport.hooks(config, chrome_driver)
 
-  config.filter_run_excluding(twilio_verify_specs: true) unless RadConfig.twilio_verify_enabled?
```
elbzero, line 136: I can't remember removing these, did you do this @garyfoster ?

spec/services/sendgrid_status_receiver_spec.rb
```diff
@@ -93,7 +93,7 @@
       end
     end
 
-    context 'with expired user', :user_expirable_specs do
```
elbzero, line 96: same here

spec/shared/models/user_spec.rb
```diff
@@ -134,9 +134,8 @@
   describe 'validate email address' do
     before { Company.main.update! valid_user_domains: %w[example.com rubygems.org] }
 
-    it 'rejects unauthorized email addresses' do
+    it 'rejects unauthorized email addresses', :valid_user_domain_specs do
```
elbzero, line 137: and here

spec/shared/models/user_spec.rb
```diff
@@ -180,9 +179,8 @@
 
     before { Company.main.update! valid_user_domains: %w[example.com rubygems.org] }
 
-    it 'rejects unauthorized email addresses' do
+    it 'rejects unauthorized email addresses', :valid_user_domain_specs do
```
elbzero, line 182: and here

spec/shared/system/invitations_spec.rb
```diff
@@ -133,7 +133,7 @@
           expect(page).to have_content "Last name can't be blank"
         end
 
-        it 'because of invalid email' do
+        it 'because of invalid email', :valid_user_domain_specs do
```
elbzero, line 136: and here

spec/rails_helper.rb
garyfoster: I removed them all as they are not needed, they're only needed in the shared version of rails_helper that gets copied into each project

PR #1167, opened 2025-07-16
Marketing Site Updates

calebtocco:
Task 10160

### DIFF
--- a/app/helpers/rad_common/application_helper.rb
+++ b/app/helpers/rad_common/application_helper.rb
@@ -26,7 +26,7 @@ module RadCommon
 
     def avatar_image(user, size)
       if RadConfig.avatar? && user.avatar.attached?
-        image_tag(user.avatar.variant(resize: '50x50'))
+        image_tag(user.avatar.variant(resize_to_limit: [50, 50]))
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -0,0 +1,26 @@
+class ApplicationController < ActionController::Base
```
garyfoster, line 1: this was moved from the radstack project, I confirmed all other apps had the minimal version of this file without the redirect_root_domain action, which isn't active w/o the marketing site config setting

app/controllers/pages_controller.rb
```diff
@@ -27,6 +27,6 @@ def privacy
     def marketing_site?
       return true if RadConfig.force_marketing_site?
 
-      RadConfig.allow_marketing_site? && request.host.start_with?('www.')
+      RadConfig.allow_marketing_site? && request.host.start_with?("#{RadConfig.marketing_subdomain}.")
```
garyfoster, line 30: we're using wwwtest for the PUG for now, could probably remove this config setting once we switch that to www

PR #1181, opened 2025-07-31
System Usage Trend Graphs

vincentdinic:
Task 9348

### DIFF
--- a/spec/dummy/package.json
+++ b/spec/dummy/package.json
@@ -14,6 +14,7 @@
     "@rails/ujs": "^6.0.5",
     "@sentry/browser": "^8.38.0",
     "bootstrap": "^5.3.7",
+    "chartkick": "^5.0.1",
     "esbuild": "^0.25.8",
     "fullcalendar": "^6.1.11",
     "jquery": "^3.3.1",
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -596,8 +596,7 @@ DEPENDENCIES
   bootsnap
   capybara
   capybara-selenium
-  devise-twilio-verify!
-  haml_lint (= 0.55.0)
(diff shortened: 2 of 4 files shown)


vincentdinic:
I just put the gem chartkick and the js import inside rad_common for initial implementation, if you like it, we can move that to rad_common_js, as mayoshifts also uses chartkick and chart.js for their user dashboards, so we can extract that.

app/views/system_usages/index.html.haml
```diff
@@ -14,11 +14,24 @@
...
           %tr
             %td= @system_usage_search.usage_data[item_index].first[:name]
             %td= @system_usage_search.usage_data[item_index].first[:current_count]
-            - @system_usage_search.date_column_ranges.each_with_index do |_header, header_index|
+            - @system_usage_search.date_column_ranges.each_with_index do |header, header_index|
               %td= @system_usage_search.usage_data[item_index][header_index][:value]
+              - data << [header[:label], @system_usage_search.usage_data[item_index][header_index][:value]]
             %td= @system_usage_search.total item_index
+            %td
+              // Note: In BS5 this will need to change to add d-grid, remove btn-block and use bs_toggle and bs_target
```
garyfoster, line 28: are these comments applicable still?

spec/dummy/package.json
```diff
@@ -6,6 +6,8 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "^6.1.1",
+    "chart.js": "^4.5.0",
```
garyfoster, line 9: check with @calebtocco, not sure if this is supposed to be in here or in rad_common_js, and just to confirm, do we need both the gem and these added?

PR #1229, opened 2025-09-30
replace legacy twilio verify gem with devise-two-factor

garyfoster:


app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Users
+  class SessionsController < Devise::SessionsController
+    def create
+      user_params = params.require(:user).permit(:email, :password, :remember_me)
```
garyfoster: Claude created this, we should carefully review and vet it since it involves security

app/controllers/users/two_factor_auth_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+        response.status == 'pending'
+      rescue StandardError => e
+        Rails.logger.error "Failed to send SMS: #{e.message}"
+        false
+      end
+
+      def verify_sms_code
+        return false unless @user.mobile_phone.present? && params[:otp_attempt].present?
+
+        response = TwilioVerifyService.verify_sms_token(@user.mobile_phone, params[:otp_attempt])
```
garyfoster: I'm not sure if we should still use TwilioVerify but it seems to work

PR #1464, opened 2026-04-15
exclude more tables in minimal clone local

garyfoster:
Task 13720

### DIFF
--- a/app/services/heroku_commands.rb
+++ b/app/services/heroku_commands.rb
@@ -6,7 +6,7 @@ class HerokuCommands
   EXCLUDED_TABLES = {
     'full' => [],
     'exclude_audits' => %w[audits],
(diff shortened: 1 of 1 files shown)


app/services/heroku_commands.rb
```diff
@@ -6,7 +6,7 @@ class HerokuCommands
   EXCLUDED_TABLES = {
     'full' => [],
     'exclude_audits' => %w[audits],
-    'minimal' => %w[audits contact_logs contact_log_recipients embeddings]
+    'minimal' => %w[active_storage_attachments active_storage_blobs audits contact_logs contact_log_recipients embeddings login_activities]
```
garyfoster, line 9: need to test this on some apps and also make sure it works in the non pull clone method

